PR #65, opened 2022-01-20
append signature

Nikbert:
Pulseq Version 1.4 requires a md5_hash signature at the end of the .seq file

This is a function that appends the signature to make the .seq file compatible with Pulseq V1.4.
Calling the function "add_signature.add_md5_sig(file_name)" after the "seq.write()" function will append the required signature.

Describe alternatives you've considered:
An alternative would be to include the function in the seq.write function.
This would be the preferred solution in the future.


sravan953:
Thank you for the PR + issue (#65 ) @Nikbert! :)
Commit f36f91a7dd7ff7c0b37ce0f559312c918f63afe4 should fix this.

PR #66, opened 2022-01-24
Open sequence output file in "with" context to be sure the file is closed

mavel101:
Currently, the output sequence is not closed at all. This commit puts the open statement in a with context to be sure, the file is closed properly.


sravan953:
@mavel101 I'm confused by what Github is showing me: "mavel101 wants to merge 2 commits into imr-framework:dev from mavel101:dev"?

PR #67, opened 2022-02-10
Fix a bug, that returns the wrong trigger channel in get_block

mavel101:
This commit fixes a bug, where the wrong trigger channel is returned or get_block might even fail completely for the "ext1" trigger channel.


mavel101:
Yes, it is still not working. You can reproduce it by adding one trigger to a sequence and getting that block:
```
seq = Sequence()
trig =  make_digital_output_pulse(channel='ext1', duration=10e-6, delay=0)
seq.add_block(trig)
seq.get_block(1)
```
This will fail in the current code, as the channel index is wrong. This is fixed in the pull request.

PR #74, opened 2022-07-21
Fix missing install module for version.py

bilal-tasdelen:
version.py was not added as an install target and not copied into install directory. This causes an exception when imported from sequence.py. This commit adds it as py_module to setup.py.


sravan953 (approved):
Thanks for catching this 👍🏼

PR #75, opened 2022-07-21
Fix the magnitude slew rate calculation in the report.

bilal-tasdelen:
Some arrays were incorrectly sliced, which caused slew rate to be calculated incorrectly.

Also fixes a typo in units.


sravan953 (approved):
Thanks for the fixes!

PR #76, opened 2022-07-21
Fix time unit not displaying correctly, and add gradient unit selection.

bilal-tasdelen:
Displaying gradients in terms of mT/m is more inline with convention.


sravan953 (approved):
Thanks for the fix + feature enhancement!

PR #78, opened 2022-07-28
Add missing paranthesis in get_block() for arb grad times.

bilal-tasdelen:
Missing parenthesis in `get_block()` caused incorrect timescale of arbitrary gradients.


sravan953:
Thank you!

PR #79, opened 2022-07-28
Fix incorrect array creation at waveform_and_gradients().

bilal-tasdelen:
The way extended trapezoids are added to the waveforms array was incorrect due to the outermost list not being numpy array, (thus, not having `.shape` property), and being wrapped in an additional list, which caused the shape to be (2, 1, N) instead of (2, N), where N is the number of waveform points. Both of these bugs caused separate exceptions.


sravan953:
Thank you!

PR #80, opened 2022-07-28
Fix missing paranthesis at per axis slew rate calc.

bilal-tasdelen:
There was one more missing parenthesis to fix the slew rate calculation at test report.


sravan953:
Thanks!

PR #81, opened 2022-07-28
[Feature Suggestion] Option to call plt.show() later for plotting.

bilal-tasdelen:
This PR adds an optional input parameter to seq.plot() that controls if plt.show() is to be called by the function or by the user. This is useful for Matplotlib, since, once show() is called, the rest of the script will be blocked till the plots are closed, meaning, no further modification on the plots can be done by the user.

A very convenient use case for me is adding data cursors, similar to Matlab's data tips using the [mplcursors](https://mplcursors.readthedocs.io/en/stable/index.html) library.


sravan953:
Thank you!

PR #83, opened 2022-08-04
fix calculation of bandwidth of adiabatic hypsec pulse

schuenke:
When trying to use a slice selective adiabatic hypsec pulse, I got several errors related to the calculation of the bandwidth of the hypsec pulse. I fixed the obvious ones like a wrong import of _calc_rf_center()_ in _calc_rf_bandwidth()_ and a wrong attribute name (rf.sig instead of rf.signal) and provide a new solution for the ___find_flank()_ function:

```python
    def __find_flank(x, f, c):
       m = np.max(np.abs(f))
       f = np.abs(f) / m
       i = np.argwhere(f > c)[0]

       return x[i]
```

Using calc_rf_bandwidth() with a cut_off value of 0.5 instead of 0.1 (which makes sense imo because the signal gets normalized to 1 and we are interested in the width at half maximum ?!) yields a bw of 1280 Hz, which is close to the literaure value of ~ 1250 Hz for the hypsec pulse with default settings (beta = 800, mu =4.9)

HOWEVER, simply using the analytic expression for the bw of a hypsec pulse makes far more sense imo. It is given by 

**bw = mu * beta / pi** 

As a reference, here the passage from Bernstein - Handbook of MRI pulse sequences:

grafik

Using the analytic expression is the way I implemented it now in _make_adiabatic_pulse.py_

### DIFF
--- a/pypulseq/calc_rf_bandwidth.py
+++ b/pypulseq/calc_rf_bandwidth.py
@@ -3,7 +3,7 @@ from typing import Union, Tuple
 
 import numpy as np
 
-from pypulseq import calc_rf_center
+from pypulseq.calc_rf_center import calc_rf_center
 
 
 def calc_rf_bandwidth(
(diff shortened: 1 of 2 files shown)


bilal-tasdelen:
@sravan953 I agree with @schuenke. I tested it, and it seems to give the correct bandwidth. It is much cleaner and less hand-wavy.

PR #86, opened 2022-08-20
Add missing input set case for trapezoids.

bilal-tasdelen:
There is a missing case for defining trapezoids, when rise time, flat time and area is given. It causes the amplitude to be calculated as flat_area/flat_time, which evaluates to 0 and causes trapezoid to be a 0 area trapezoid. This commit adds the missing case.

### DIFF
--- a/pypulseq/make_trapezoid.py
+++ b/pypulseq/make_trapezoid.py
@@ -84,6 +84,8 @@ def make_trapezoid(
     if flat_time != -1:
         if amplitude != 0:
             amplitude2 = amplitude
+        elif (area != 0) and (rise_time > 0): # We have rise_time, flat_time and area.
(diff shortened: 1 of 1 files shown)


sravan953:
Thanks for the PR! What use case would this address?

bilal-tasdelen:
It will address the case where we want two gradients to have the exact same timing (rise time and flat time), but different areas.

A more specific use case that I encountered: I wanted to design bridged spoilers for spin echo, that have the same timing, but different areas (4pi and 8pi), in order to cycle between different spoilers so that we can avoid unwanted stimulated echoes. Having the spoilers with exact same timing simplifies the bridging and sequence timing.

bilal-tasdelen:
PR had a small bug causing some input sets to throw an exception, but it's fixed and tested now.

PR #84, opened 2022-08-09
Gradient waveforms fix

bilal-tasdelen:
Tries to fix #72.

PR #87, opened 2022-08-21
Fix side effects of using numpy.round

bilal-tasdelen:
This commit fixes the bugs introduced in #84 due to numpy's rounding function.
np.round outputs the value as np.float64, which can not be automatically cast as int, thus, can not be used for slicing the arrays.


sravan953:
Won't a simple `int()` casting work?

bilal-tasdelen:
I think it is safe to assume `waveforms_and_times()` will return time points on gradient raster time. So you are right, it is enough.

PR #100, opened 2023-02-09
np.int is deprecated in numpy>=1.20.0

skarrea:
With the current requirements in [setup.py](https://github.com/imr-framework/pypulseq/blob/dev/setup.py#L37) numpy >=1.20 will be installed where `np.int` is deprecated.

Changing all occurrences of `np.int` to `np.int32` is compatible with both `1.19.5 <= numpy <= 1.20.0`, and `numpy >= 1.20.0` (tested with 1.23.5). `np.int` can also be changed to `np.int64` if necessary.

### DIFF
--- a/pypulseq/Sequence/block.py
+++ b/pypulseq/Sequence/block.py
@@ -38,7 +38,7 @@ def set_block(self, block_index: int, *args: SimpleNamespace) -> None:
         If a gradient that doesn't end at zero is not aligned to the block boundary.
     """
     events = block_to_events(*args)
-    self.block_events[block_index] = np.zeros(7, dtype=np.int)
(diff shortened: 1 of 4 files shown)


sravan953:
Thanks for catching this! I fixed it in one of the recent commits (3a397e4).

skarrea:
Not really. I would guess it comes with a minimal performance penalty since `int` will be interpreted as `np.int_` (since numpy arrays can't contain `int`). `np.int_` will then default to `np.int32` or `np.int64` depending on the system.

sravan953:
You're right; it's best to avoid that penalty by directly specifying `np.int32` ourselves.

sravan953 (approved):

Issue #106, opened 2023-03-13
Issue adding new shape


tblazey:
Hi,

I've been having a problem adding rf blocks with different phase modulation to my sequence. I can add the first couple of pulses, but after that I see repeated shapes when I plot the sequence. After a bit of searching, I think I have narrowed it down to the find_or_insert function in event_lib.py. With my shapes, find_or_insert was reporting there was an identical shape in the dictionary even when that shape wasn't present.

I believe the issue is that because my shapes are longer than numpy.get_printoptions()['threshold'], array2string is returning a shortened string with just the the first and last few elements separated by a '...'.  In my case, many of the shapes I wanted to add had the same beginning and ending values. If set the threshold parameter in array2string to np.inf, I can add the additional shapes.

Thanks,

-Tyler

https://github.com/imr-framework/pypulseq/blob/c246ff9ff0018ab9b5a6b4add3850f068f61b034/pypulseq/event_lib.py#L109-L111

sravan953:
Hello! Thanks for reporting the bug and also identifying a workaround. Could you please share code to reproduce the bug? If you're also interested in submitting a PR to fix this bug, I'd much appreciate it!

tblazey:
Hi,

An example is below. I actually had a bit of trouble reproducing the bug with generic pulses, so I have attached my pulse as well. I suspect that this issue interacts with shape compression in some way, but I'm not sure how.

```
import pypulseq as pp
import numpy as np

#Create sequence object
seq = pp.Sequence()

#Load in rf pulse
rf_data = np.loadtxt('siemens_pyr_plateau_slab_rf.csv', delimiter=',')
rf_pulse = rf_data[0, :] * np.exp(1j * rf_data[1, :])

#Loop through phases
n = 8
for i in range(n):

   #Add new RF pulse to sequence
   rf_ev = pp.make_arbitrary_rf(rf_pulse * np.exp(1j * np.pi / n * i),  np.pi / 2, system=seq.system)
   seq.add_block(rf_ev)
   
   #Print RF ID
   print(seq.block_events[i + 1][1])
```

By default, I only see one shape and the code prints '1, 1, 1, 1, 1, 1, 1, 1'. With the change in #108, it prints '1, 2, 3, 4, 5, 6, 7, 8'.

Thanks!

-Tyler

[siemens_pyr_plateau_slab_rf.csv](https://github.com/imr-framework/pypulseq/files/10991435/siemens_pyr_plateau_slab_rf.csv)

[...]

Issue #105, opened 2023-03-11
Possible bug in sequence.py


tblazey:
Hi,

I noticed that the adc times returned by waveforms_and_times in sequence.py are rather large. I think the issue starts at line 1475 of sequence.py

```
t_adc.extend(
  np.arange(block.adc.num_samples)
  + 0.5 * block.adc.dwell
  + block.adc.delay
  + curr_dur
)
```

I think the np.arange call needs to be multiplied by 0.5 * block.adc.dwell instead of added. I hope I am referencing the most current version of pypulseq. Apologies for not doing that last time.

Thanks,

-Tyler

sravan953:
Thanks for reporting this bug and also submitting a bug-fix PR!

bilal-tasdelen:
Fixed with #110.

PR #108, opened 2023-03-16
bug fixes to event_lib.py and sequence.py

tblazey:
Hi,

These are suggested fixes for the following issues #105 and #106. This is my first pull request, so apologies if I am making any mistakes.

Thanks!

-Tyler

### DIFF
--- a/pypulseq/Sequence/sequence.py
+++ b/pypulseq/Sequence/sequence.py
@@ -1473,8 +1473,8 @@ class Sequence:
 
             if block.adc is not None:  # ADC
                 t_adc.extend(
-                    np.arange(block.adc.num_samples)
-                    + 0.5 * block.adc.dwell
+                    np.arange(block.adc.num_samples) * block.adc.dwell
+                    + block.adc.dwell * 0.5
                     + block.adc.delay
                     + curr_dur
                 )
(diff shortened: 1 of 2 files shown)


bilal-tasdelen:
@tblazey Thanks for the PR. Solution to #105 seems good to me. 

@sravan953 I had a quick look into `find_or_insert()` function. It looks like it is creating a string from the shape array to use it as a dictionary key. Is my understanding correct?

If that is the key, it is very slow and error-prone. Adding `threshold=np.inf` will just mask the issue.

A better hack would be just replacing `np.array2string()` call with `new_data.to_bytes()`. This will be 1) much faster as it doesn't do any formatting, 2) precise, as it will make sure two shapes are identical byte by byte.

An even more proper solution would be calculating a hash from the array, and using it as the key.

Further info:
[https://stackoverflow.com/questions/39674863/python-alternative-for-using-numpy-array-as-key-in-dictionary](https://stackoverflow.com/questions/39674863/python-alternative-for-using-numpy-array-as-key-in-dictionary)

sravan953:
@bilal-tasdelen I was not aware of performance issues, thank you for bringing it to my notice. 

@tblazey Thank you for the PR! Always super happy to see the community participating.
I request you to submit 2 separate PRs for each of the bugs if you don't mind? For the PR fixing the bug in `event_lib.py`, could you please implement bilal's suggested fix, referring to the code example he posted in the previous comment?

tblazey:
Hi,

I wasn't sure how to split the original request, so I submitted separate pull requests (#110 and #111). Hopefully that is ok.

Thanks,

-Tyler

PR #112, opened 2023-03-27
Negative gyromagnetic ratio

Cwis:
I am working with nuclei having negative gyromagnetic ratio (such as 129mXe) and this leads to error when using pypulseq.
So far, I have identified issues in making trapezoidal gradients and in Sequence/Block.py. In those files I had to add an abs() on the system max_grad and max_slew. This might be needed elsewhere if using other functions.


bilal-tasdelen:
Hi @Cwis, thanks for raising this issue. Excuse my ignorance, but this is the first time I hear about negative gyromagnetic ratio. 

Can you tell us the implications of this, so we can come up with a plan to how to support it properly? 

The only implication I can think of is the precession direction. If that is the case, would simply giving the absolute value of the gyromagnetic ratio, rather than the negative value, solve the issue?

curtcorum:
Here is the basic discussion:
https://en.wikipedia.org/wiki/Gyromagnetic_ratio#For_a_nucleus

"Spin Dynamics" has the most detailed discussion of sign conventions for NMR and MRI and the fact that they are not always consistent in the literature:
Levitt, M. H. (2013). Spin Dynamics: Basics of Nuclear Magnetic Resonance. Germany: Wiley.

curtcorum:
The precession direction also requires that the circular polarized mode needs to be adjusted (reversed) for the transmit coil. This is no doubt platform dependent for how it gets set, either by selection of the nuclei of interest or other switches.

bilal-tasdelen:
@Cwis Thanks for sharing the extra information about negative gyromagnetic ratio.

As you said, I don't think currently there is any mechanism in Pulseq that allows you to change reverse CP mode for transmit coil. That, most likely, requires a change in the interpreter code. For that, you can create an issue on Pulseq GitHub: [https://github.com/pulseq/pulseq](https://github.com/pulseq/pulseq)

It still looks like, the rest can be fixed by giving the absolute value of the negative gyromagnetic ratio to the PyPulseq. I think it is a better approach, rather than taking absolute values for `max_slew` and `max_grad`, those values should never be negative in the first place.

Let me know if my suggestion does not solve your issue, or misses some points.

Cwis:
@bilal-tasdelen Thank you for your answer, I am working on simulations with nuclei that have negative gyromagnetic ratio (metastable Xenon isomers). It happens when the magnetic moment and the angular moment of the nuclei are in opposite direction (but still co-aligned). It's true that the only difference is the precession direction - which is important to my study - but it should not change anything for the sequence design with PyPulseq, does it ?
If not, then I will just use the absolute value when designing the sequence.

bilal-tasdelen:
@Cwis That is my understanding as well. I will close this for now, but feel free to reopen if there is any unforeseen consequence of negative gyromagnetic ratio in the sequence design.

PR #113, opened 2023-04-03
latest make_sinc_pulse dont return gradient by default

yupbank:


bilal-tasdelen:
@yupbank Thanks for catching this!

PR #114, opened 2023-04-24
Hacky fix for waveforms_and_times rf append floating point issue.

btasdelen:
When we have RF pulses with non-zero start/end times, waveforms_and_times() function tries to add trailing/leading zeros as necessary to the beginning and end of the RF waveform. To avoid having duplicate times, leading zero is added at time `t[0]-eps,` trailing zero is added to the `t[-1]+eps`, where `eps` is supposedly "smallest floating increment". These are all inline with upstream MATLAB `Pulseq` implementation.

Trouble begins with the `eps`. `eps` is not actually the smallest increment. Floating point have variable increments depending on the number of significant digits. Quoting `numpy` documentation:

> 
> eps: float
> 
> The difference between 1.0 and the next smallest representable float larger than 1.0. For example, for 64-bit binary floats in the IEEE-754 standard, eps = 2**-52, approximately 2.22e-16.

It is only true if we are close to 1.0. Time vector is an unbounded vector, so our comparison with `eps` becomes incorrect as we add `eps` to larger and larger floats. When we apply `np.unique()` to these values, they are identified as duplicates, thus we raise:

```python
raise Warning(
                           "Not all elements of the generated time vector are unique."
                        )
```

I could not find a solution to fix this without changing the way we look at things here, so I just added `50*eps` arbitrarily to silence the error. I am open to suggestions on how to solve this properly. I am also not sure why we need to add trailing or leading zeros, maybe the answer lies removing that.

--------

The second issue this PR addresses is, using `np.unique` is extremely slow, and prohibits the use of the function for large sequences. It seems like it is used as a check to see if we have repeating time points. I would say it is an overkill. We do not expect to have an arbitrarily ordered time vector. If there is a repeat, it is most likely happening at consecutive time points. Thus, I replaced `np.unique` with a call to `np.diff` and a comparison with `eps`. This will make sure that time points are monotonically increasing. This way, I observed around 2.7 times improvement in speed. `np.diff` is still slow, in the near future, a better approach is needed.

Sorry for the long post, let me know what you think.

### DIFF
--- a/pypulseq/Sequence/sequence.py
+++ b/pypulseq/Sequence/sequence.py
@@ -1447,6 +1447,7 @@ class Sequence:
                         [curr_dur + t, block.rf.freq_offset, block.rf.phase_offset]
                     )
                 if append_RF:
+
                     rf_piece = np.array(
                         [
                             curr_dur + rf.delay + rf.t,
@@ -1460,12 +1461,12 @@ class Sequence:
                     out_len[-1] += len(rf.t)
 
                     if np.abs(rf.signal[0]) > 0:
-                        pre = np.array([[curr_dur + rf.delay + rf.t[0] - eps], [0]])
+                        pre = np.array([[rf_piece[0, 0] - 50*eps], [0]], dtype=np.complex128)
                         rf_piece = np.hstack((pre, rf_piece))
                         out_len[-1] += pre.shape[1]
 
                     if np.abs(rf.signal[-1]) > 0:
-                        post = np.array([[curr_dur + rf.delay + rf.t[-1] + eps], [0]])
+                        post = np.array([[rf_piece[0, -1] + 50*eps], [0]], dtype=np.complex128)
                         rf_piece = np.hstack((rf_piece, post))
                         out_len[-1] += post.shape[1]
 
@@ -1510,10 +1511,14 @@ class Sequence:
                             :, wave_cnt[j] + np.arange(length - 1)
                         ] = wave_data_local[:, 1:]
                         wave_cnt[j] += length - 1
-                    if wave_cnt[j] != len(np.unique(wave_data[j][0, : wave_cnt[j]])):
+
+                    rftdiff = np.diff(wave_data[j][0, : wave_cnt[j]])
+                    if np.any(rftdiff < eps):
                         raise Warning(
-                            "Not all elements of the generated time vector are unique."
+                            "Time vector elements are not monotonically increasing."
                         )
+
+
 
         # Trim output data
         for j in range(shape_channels):
--- a/pypulseq/Sequence/sequence.py
+++ b/pypulseq/Sequence/sequence.py
@@ -1461,12 +1461,12 @@ class Sequence:
                     out_len[-1] += len(rf.t)
 
                     if np.abs(rf.signal[0]) > 0:
-                        pre = np.array([[rf_piece[0, 0] - 50*eps], [0]], dtype=np.complex128)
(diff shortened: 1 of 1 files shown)


pypulseq/Sequence/sequence.py
```diff
@@ -1510,11 +1511,15 @@ def waveforms_and_times(
                             :, wave_cnt[j] + np.arange(length - 1)
                         ] = wave_data_local[:, 1:]
                         wave_cnt[j] += length - 1
-                    if wave_cnt[j] != len(np.unique(wave_data[j][0, : wave_cnt[j]])):
+
+                    rftdiff = np.diff(wave_data[j][0, : wave_cnt[j]])
```
sravan953, line 1515: Request you to try `list(set(...))`, is that faster?

sravan953 (changes requested):

pypulseq/Sequence/sequence.py
```diff
@@ -1460,12 +1461,12 @@ def waveforms_and_times(
                     out_len[-1] += len(rf.t)
 
                     if np.abs(rf.signal[0]) > 0:
-                        pre = np.array([[curr_dur + rf.delay + rf.t[0] - eps], [0]])
+                        pre = np.array([[rf_piece[0, 0] - 50*eps], [0]], dtype=np.complex128)
```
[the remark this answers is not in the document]
bilal-tasdelen: 1. `rf_piece` is a numpy array, so we can't unpack it with * operator.
2. 50 is an arbitrary number that worked for my case, longer sequences fail even with 50. I tried what you suggest, but eventually, this causes the added time point to have a larger step than raster time, so we have negative time jumps.

Maybe, I should just add, 0.1*rf raster time? My understanding is, we want the time difference to be as small as possible, because, let's say, we integrate the RF or try to do Bloch simulation, adding this point will change the result. Maybe I should ask Dr. Maxim why he added this in the first place.

3. For that line, I don't have to specify, as rf_piece is already complex128. I can remove that.

pypulseq/Sequence/sequence.py
bilal-tasdelen: I tried that, it is extremely slow.

As an aside, I think we should not do this check here, but rather during RF creation, for the following reasons:

1. It does not make sense to for the sequence developer to create an RF object with non-monotonically increasing time points in the first place. `waveforms_and_times()` function may not be called typically, so this issue may not be caught in this case.

2. We create RF object once, and add it many times (i.e. every TR). This way, we are checking the same RF pulse many times unnecessarily, hence the slow-down.

I can also ask Maxim about this design choice. But if you are convinced with my argument, I can remove this check and put it inside make RF functions.

bilal-tasdelen: Discussion with Maxim: [https://github.com/pulseq/pulseq/issues/25](https://github.com/pulseq/pulseq/issues/25)

TL;DR: He is planning to make some changes to reduce the calls to unique(). Till then, I am leaving it at diff.

pypulseq/Sequence/sequence.py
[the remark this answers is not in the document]
bilal-tasdelen: Discussion with Maxim: [https://github.com/pulseq/pulseq/issues/25](https://github.com/pulseq/pulseq/issues/25)

As suspected, proper handling of this is a bit involved. I will use 0.1*rf_raster_time for now, and we can see if it creates a problem later on.

sravan953 (approved):

PR #125, opened 2023-07-20
Bug fixes that produced incorrect test reports

FrankZijlstra:
Fixed several bugs in Sequence and test_report that result in incorrect test reports (e.g. wrong slew rates!) and incorrect k-space trajectory calculation.

Checked test reports for gre, tse, gre_radial, ute, and epi_se_rs to be identical with MATLAB.

Additionally made some minor changes that avoid some warnings (bad conditioning in polynomial fit and division by zero).

### DIFF
--- a/pypulseq/Sequence/ext_test_report.py
+++ b/pypulseq/Sequence/ext_test_report.py
@@ -44,7 +44,7 @@ def ext_test_report(self) -> str:
             i2check.append(index_echo + 1)
 
         for a in range(len(i2check)):
-            v_i_to_0 = k_traj_adc[:, index_echo]
+            v_i_to_0 = -k_traj_adc[:, index_echo]
             v_i_to_t = k_traj_adc[:, i2check[a]] - k_traj_adc[:, index_echo]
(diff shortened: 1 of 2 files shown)


btasdelen:
Thank you for the PR @FrankZijlstra. That is quite a substantial fix, thanks for the effort! 

The patch for the test report looks straightforward, but I am not really familiar with the polynomial codebase. Maybe @sravan953 can comment on that?

An aside: I am planning to propose a PR to change the internal representation of time to an integer in units of nanoseconds rather than float in units of seconds. That should hopefully resolve the issues regarding empirical selection of `eps`.

zhengliuer:
> An aside: I am planning to propose a PR to change the internal representation of time to an integer in units of nanoseconds rather than float in units of seconds. That should hopefully resolve the issues regarding empirical selection of `eps`.
This is really a great idea!

sravan953:
Hi all, apologies for the delay. I'll look into this in a day or two! Thank you for the PR @FrankZijlstra

PR #126, opened 2023-07-22
codespell: add config and action to codespell the code to avoid known typos

yarikoptic:


sravan953:
Thanks for this @yarikoptic ! Does codespell then submit a PR with corrections?

yarikoptic:
no -- just fails the workflow and it would be for contributor to fix locally (by running codespell or manually)

schuenke:
@yarikoptic We introduced typos as a pre-commit hook in #207 so I think we can close this, no?

schuenke:
> @yarikoptic We introduced typos as a pre-commit hook in #207 so I think we can close this, no?

PR #131, opened 2023-08-21
Added PNS calculations

FrankZijlstra:
This is a direct translation of the MATLAB Pulseq `calcPNS` function to python, renamed to `calculate_pns`. This includes a partial translation of the MATLAB safe_pns_prediction repository (https://github.com/filip-szczepankiewicz/safe_pns_prediction), put into utils/safe_pns_prediction.py. To test the functionality, this adds the pnsTest sequence "example" from MATLAB Pulseq.

A few, mostly stylistic, things may need some review:
- `readasc` and `asc_to_hw` are Siemens specific, might need to be put in a place for Siemens-specific code? (in Pulseq only `readasc` is separated)
- All functions from the SAFE code are in the same file, and maintain the MATLAB comments (as far as I'm concerned it really is not meant to be more than a direct translation). And what about license terms: safe_pns_prediction is BSD-3, pypulseq is AGPL-3?
- Location of the SAFE code (`/utils`) and `pns_test.py` (sequence examples)?

### DIFF
--- a/pypulseq/Sequence/calc_pns.py
+++ b/pypulseq/Sequence/calc_pns.py
@@ -1,0 +1,99 @@
+from types import SimpleNamespace
+from typing import Tuple
+import matplotlib.pyplot as plt
+import pypulseq as pp
+import numpy as np
+
+from pypulseq import Sequence
+from pypulseq.utils.safe_pns_prediction import safe_gwf_to_pns, safe_plot
+
+from pypulseq.utils.siemens.readasc import readasc
+from pypulseq.utils.siemens.asc_to_hw import asc_to_hw
+
+
+def calc_pns(
+        obj : Sequence, hardware : SimpleNamespace, do_plots: bool = True
+        ) -> Tuple[bool, np.array, np.ndarray, np.array]:
+    """
+    Calculate PNS using safe model implementation by Szczepankiewicz and Witzel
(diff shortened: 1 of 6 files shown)


btasdelen (changes requested):
To address your questions:
- I agree, it is better to Siemens specific functions into its own directory. If you are positive read_asc is explicitly Siemens specific, we don't have to follow Pulseq hierarchy, we can also put it in Siemens.
- BSD is compatible with AGPL. I am not clear if it is a must, as you did modify the code significantly even though it is a translation, but we should retain the LICENSE and copyright notice for it. There are a couple of ways to do it:
1. We can paste their license file into the file `safe_pns_prediction`.
2. We can append our license with their license.
3. We can paste the license and explain the licensing situation in the documentation.

It is your call to pick, but 1 seems the easiest.
 
 - I think the file positions are ok, but I would consider putting `calc_pns` into `utils` as well. Your call.

FrankZijlstra:
I moved the Siemens functions into `utils/siemens` and added the BSD license into `safe_pns_prediction`.
I did not move `calc_pns` to remain consistent with MATLAB pulseq (similar as for `ext_test_report`, `read_seq`, etc.)

I updated the previous commit to keep the pull-request to a single commit.

PR #134, opened 2023-09-19
Remove sigpy dependency and add the files defining RF directly.

btasdelen:
This PR aims to remove SigPy dependency. 

What works: `make_adiabatic_pulse` seems to be working.

What is missing: `make_sigpy_pulse` still requires `sigpy`. The reason is, it uses plot functions from `sigpy`, and it requires a much larger part of the library. If those plot lines are replaced or removed, we can completely eliminate the dependency.

### DIFF
--- a/pypulseq/make_adiabatic_pulse.py
+++ b/pypulseq/make_adiabatic_pulse.py
@@ -2,7 +2,7 @@ from types import SimpleNamespace
 from typing import Tuple, Union
 
 import numpy as np
-from sigpy.mri.rf import hypsec, wurst
+from pypulseq.utils.sigpy.adiabatic import hypsec, wurst
 
 from pypulseq import eps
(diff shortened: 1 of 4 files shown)


schuenke:
I am not 100% convinced that this is the way to go. 

As you describe, `make_sigpy_pulse` uses a larger part of the _sigpy_ library and if we remove this additional functionality (e.g. the plot methods), we will break the backwards compatibility. Thus, we either have to include all the functionality into pypulseq, as well, or we would need to change the major version of pypulseq if we want to stick with [semantic versioning](https://semver.org/).

Therefore, I would still suggest that we temporarily change the dependency in our dev branch to `git+https://github.com/mikgroup/sigpy.git` and change it back to `sigpy>0.1.25` as soon as the sigpy developers release their fix (will probably be version 0.1.26).

PR #140, opened 2023-10-23
Reworked piecewise polynomial implementation in Sequence to use scipy's PPoly

### LINKED ISSUE #107: 'Sequence' object has no attribute 'gradient_waveforms' in Sequence.calculate_kspace
Hi,

https://github.com/imr-framework/pypulseq/blob/a18e34f89bc1d9a4a30e467013be00a8f17a3439/pypulseq/Sequence/sequence.py#L214
results in '`Sequence' object has no attribute 'gradient_waveforms'`
as `gradient_waveforms` has been removed in https://github.com/imr-framework/pypulseq/commit/22fb57c2006208510d0fd573fc6757d71fbae4c8

Not sure why as I am not familiar with  thecodebase and the commit message does not really explain it... 

Cheers,
Felix

FrankZijlstra:
This implements `calculate_kspacePP` using scipy's `PPoly` object, which behaves very similar to the Matlab implementation. This is obviously much cleaner, and also much faster. A helper function `get_gradients` is also available to immediately get access to the waveforms in `PPoly` format.
I renamed the function to `calculate_kspace`, because this function was no longer functional, and not intended to be used anymore anyway. I added a deprecation warning for `calculate_kspacePP`, but this function could also just be removed altogether. This also resolves #107.

Furthermore, I split `waveforms_and_times` into three functions: `waveforms`, `adc_times`, and `rf_times`. This allows efficient access to for example only the excitation times (e.g. when only calculating TE/TR). `waveforms_and_times` remains compatible with the previous implementation, but `rf_times` returns excitation/refocusing times and frequency/phase in separate arrays (instead of the non-intuitive `tfp_` variables).

Further small changes:
- Improved performance of calculating of `tc` in `calculate_kspace` (especially in sequences with ramp-sampling)
- Changed `block_durations` in Sequence to be a dictionary, to avoid potential issues when using non-incremental block IDs (previous code assumed this was the case)
- Fixed `waveforms_export` (though this function seems a bit superfluous with the other ways to access this information)

### DIFF
--- a/pypulseq/Sequence/block.py
+++ b/pypulseq/Sequence/block.py
@@ -225,7 +225,7 @@ def set_block(self, block_index: int, *args: SimpleNamespace) -> None:
                 "A gradient that doesn't end at zero needs to be aligned to the block boundary."
             )
 
-    self.block_durations.append(float(duration))
+    self.block_durations[block_index] = float(duration)
 
 
 def get_block(self, block_index: int) -> SimpleNamespace:
@@ -398,7 +398,7 @@ def get_block(self, block_index: int) -> SimpleNamespace:
 
             next_ext_id = ext_data[2]
 
-    block.block_duration = self.block_durations[block_index - 1]
+    block.block_duration = self.block_durations[block_index]
 
     # Enter block into the block cache
     if self.use_block_cache:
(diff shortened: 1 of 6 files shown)


sravan953:
Thanks for this! I spent a little time on the calculate k-space function but could not get the Scipy version to behave as expected. I ended up translating the MATLAB function to make it work, but that ended up being slow. Glad you got this to work!

PR #141, opened 2023-10-23
Fixed `set_block` to give errors when multiple events of the same type are specified

FrankZijlstra:
As title. There are two options here:
1. Only first commit: Give errors for all event types
2. Both commits: Give errors for RF/ADC events, but automatically add gradients together

Personally I'm in favour of option 2, because it seems like pretty natural behaviour, and it can come up in some instances: e.g. specifying a slab-selection refocusing gradient along with a phase-encoding gradient in 3D imaging. An argument against this behaviour is that it hides something the user should ideally be aware of, and you may end up with weird gradient shapes if the timing of gradients is not aligned.

### DIFF
--- a/pypulseq/Sequence/block.py
+++ b/pypulseq/Sequence/block.py
@@ -48,6 +48,9 @@ def set_block(self, block_index: int, *args: SimpleNamespace) -> None:
     for event in events:
         if not isinstance(event, float):  # If event is not a block duration
             if event.type == "rf":
+                if self.block_events[block_index][1] != 0:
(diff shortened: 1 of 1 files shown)


sravan953:
Thanks for the PR!

I'm in favour of option 1 because of the disadvantage with option 2 you mentioned. It seems like coming across an error about multiple gradients and then manually adding gradients is safer than passing multiple gradients and spending time debugging because the output looks different.

btasdelen:
I feel like there is no disagreement on the basic fix, so merging.

Issue #145, opened 2023-11-01
Cannot generate trapezoidal gradient only specifying amplitude and duration.


wtclarke:
**Describe the bug**
If you run `make_trapezoid(channel='x',  amplitude=1, duration=1)` (ignore the not sensible values) an unhandled exception is thrown.

```
            if rise_time == 0:
                rise_time = (
>                   math.ceil(abs(amplitude2) / max_slew / system.grad_raster_time)
                    * system.grad_raster_time
                )
E               UnboundLocalError: cannot access local variable 'amplitude2' where it is not associated with a value

pypulseq/make_trapezoid.py:174: UnboundLocalError

```

This is because within the if statement `elif duration > 0:` on [line 146](https://github.com/imr-framework/pypulseq/blob/dev/pypulseq/make_trapezoid.py#L146) there is a case for `amplitude == 0` [line 147](https://github.com/imr-framework/pypulseq/blob/dev/pypulseq/make_trapezoid.py#L147) but not an `else` ( for `amplitude !=0`). Within this statement `amplitude2` is declared.

**To Reproduce**
- Run `make_trapezoid(channel='x',  amplitude=1, duration=1)`
- 
**Expected behavior**
It's indicated in the documentation that this should work and it isn't handled explicitly.

**Additional context**
I'll make a PR with a fix and some tests.

wtclarke:
Fixed with the merge of #146

PR #146, opened 2023-11-01
BF: make_trapezoid with amplitude + duration as inputs.

wtclarke:
This should fix the issue described in #145. It also adds partial tests for the `make_trapezoid` function. Finally it removes a unused inport that was throwing warnings with tests in the `pypulseq/SAR/SAR_calc.py` module.

### DIFF
--- a/pypulseq/SAR/SAR_calc.py
+++ b/pypulseq/SAR/SAR_calc.py
@@ -5,7 +5,6 @@ from typing import Union
 
 import matplotlib.pyplot as plt
 import numpy as np
-import numpy.matlib
 import scipy.io as sio
 from scipy import interpolate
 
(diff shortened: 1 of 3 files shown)


wtclarke:
Before this is merged, ti would be great to understand exactly what the calling cases should be so that we can write some proper tests for them.

btasdelen:
@wtclarke I checked the PR, and the use cases you documented covers pretty much all the "useful" and well-defined cases I can think of.

```
    The user must supply as a minimum one of the following sets:
    - area
    - amplitude and duration
    - flat_time and flat_area
    - flat_time and amplitude
    - flat_time, area and rise_time
```

The complexity of input parameter sets bothers me, though. It is error-prone and not really Pythonic. I know it is something inherited from the Matlab Pulseq, so I don't know how to solve it without breaking the familiarity.

Are you considering adding test cases, or should I go ahead and merge?

wtclarke:
If you are happy with the use cases then go ahead and merge. These are all already tested for in `test_make_trapezoid.py`. There's still lots that should be added to that test set, but this is a start.

Issue #119, opened 2023-05-16
Sigpy==0.1.23 incompatible with numpy>1.20


wtclarke:
Hi,

I tried installing this package by cloning and then running `pip install .`, but when running one of the demo scripts `python pypulseq/seq_examples/scripts/write_epi_se_rs.py` the following error was returned:
```
Traceback (most recent call last):
  File "/Users/wclarke/Library/CloudStorage/OneDrive-Nexus365/projects/pps/pypulseq/pypulseq/seq_examples/scripts/write_epi_se_rs.py", line 9, in <module>
    import pypulseq as pp
  File "/Users/wclarke/miniconda3/envs/pps/lib/python3.10/site-packages/pypulseq/__init__.py", line 32, in <module>
    from pypulseq.make_adiabatic_pulse import make_adiabatic_pulse
  File "/Users/wclarke/miniconda3/envs/pps/lib/python3.10/site-packages/pypulseq/make_adiabatic_pulse.py", line 5, in <module>
    from sigpy.mri.rf import hypsec, wurst
  File "/Users/wclarke/miniconda3/envs/pps/lib/python3.10/site-packages/sigpy/__init__.py", line 12, in <module>
    from sigpy import alg, app, config, linop, prox
  File "/Users/wclarke/miniconda3/envs/pps/lib/python3.10/site-packages/sigpy/alg.py", line 8, in <module>
    from sigpy import backend, util
  File "/Users/wclarke/miniconda3/envs/pps/lib/python3.10/site-packages/sigpy/util.py", line 236, in <module>
    def dirac(shape, dtype=np.float, device=backend.cpu_device):
  File "/Users/wclarke/miniconda3/envs/pps/lib/python3.10/site-packages/numpy/__init__.py", line 305, in __getattr__
    raise AttributeError(__former_attrs__[attr])
AttributeError: module 'numpy' has no attribute 'float'.
`np.float` was a deprecated alias for the builtin `float`. To avoid this error in existing code, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations. Did you mean: 'cfloat'?
```

[...]

PR #148, opened 2023-11-09
Bump sigpy version to 0.1.26

Andrew-Dupuis:
Hi folks, SigPy looks to have merged the NumPy fixes mentioned in #119  about 3 weeks ago - this PR just bumps the version in the setup.py file. I wasn't sure about patch version updates since versioning tracks Pulseq mainline, so the only change currenlty here is in the install_requires.


btasdelen:
Thank you!

PR #158, opened 2023-12-01
Update confusing documentation for calc_duration.py

gabuzi:
This really confused me, and I got wrong results initially as I was thinking that this would sum up the durations of the provided events, while it's identifying the maximum duration.

I went ahead and wrote a bunch of tests for this, covering a wide variety of event combinations (that are most likely unrealistic if this is to be used to identify the longest event in a block). Nevertheless, they may be useful to double check the event creation functions using the `duration` args.

### DIFF
--- a/pypulseq/calc_duration.py
+++ b/pypulseq/calc_duration.py
@@ -7,7 +7,14 @@ from pypulseq.block_to_events import block_to_events
 
 def calc_duration(*args: SimpleNamespace) -> float:
     """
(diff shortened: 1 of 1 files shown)


btasdelen:
That certainly is clearer and more accurate. The way I think about it is `calc_duration` gives the `block` duration, if we are to create a block with the input events. I don't know if event duration implies for most people the delay+waveform duration, or just the waveform duration. Do you think adding a remark like "... maximum duration of its events (block duration)" makes it clearer or more confusing?

PR #162, opened 2024-01-10
Fixed `read_seq` for reading v1.3.1 sequence files

### LINKED ISSUE #161: Compatibility with Pulseq 1.3.1 broken
**Describe the bug**
Reading Pulseq 1.3.1 sequences does not work any more.

**To Reproduce**
Try to read any 1.3.1 sequence, e.g. the gre_example.seq

**Expected behavior**
PyPulseq 1.4.0 should be backwards compatible with Pulseq 1.3.1 files. The case is (supposed to be) catched in functions like Sequence.read() etc.

**Detailed description**
First of all, I get an error from https://github.com/imr-framework/pypulseq/blob/db19dda7088e91841f2e7aba34d2fdd8929ddab6/pypulseq/Sequence/read_seq.py#L224C44-L224C44, because `rf_library` has no attribute `lenghts`

As this seems not to be used, removing the line fixes this issue. Still, is `lenghts` in general supposed to be an attribute/property of `EventLibrary`? It's listed as a property [here](https://github.com/imr-framework/pypulseq/blob/db19dda7088e91841f2e7aba34d2fdd8929ddab6/pypulseq/event_lib.py#L13), but it's never defined or used?

However, the real issue IMO is the `read_seq` function, more precise [in line 254-264](https://github.com/imr-framework/pypulseq/blob/db19dda7088e91841f2e7aba34d2fdd8929ddab6/pypulseq/Sequence/read_seq.py#L254-L264).
The idea is to loop over the blocks and calculate the block durations to fill `Sequence.block_durations`, but within the `get_block` function in [line 405](https://github.com/imr-framework/pypulseq/blob/db19dda7088e91841f2e7aba34d2fdd8929ddab6/pypulseq/Sequence/block.py#L405), we (try to) set the duration of the block using the entries from `Sequence.block_durations`:

` block.block_duration = self.block_durations[block_index]`

which does not work for Pulseq 1.3.1 files, because we actually call `get_block` to calculate the block duration 

I might have a look at this myself at some point, but wanted to keep it documented already.

FrankZijlstra:
- Removed some references the to `lengths` field in `event_lib`. I removed this field in previous PRs because it was completely unused, but missed these uses.
- Fixed the handling of block durations and delays for 1.3.1 sequence files in `read_seq`.

Should fix #161

### DIFF
--- a/pypulseq/Sequence/read_seq.py
+++ b/pypulseq/Sequence/read_seq.py
@@ -216,11 +216,11 @@ def read(self, path: str, detect_rf_use: bool = False) -> None:
     if version_combined < 1004000:
         # Scan through RF objects
         for i in self.rf_library.data:
-            self.rf_library.data[i] = [
-                *self.rf_library.data[i][:3],
-                0,
-                *self.rf_library.data[i][3:],
-            ]
+            self.rf_library.update(i, None, (
+                            *self.rf_library.data[i][:3],
+                            0,
+                            *self.rf_library.data[i][3:]
+                        ))
 
         # Scan through the gradient objects and update 't'-s (trapezoids) und 'g'-s (free-shape gradients)
         for i in self.grad_library.data:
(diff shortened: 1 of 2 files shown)


pypulseq/Sequence/read_seq.py
```diff
@@ -231,36 +230,37 @@ def read(self, path: str, detect_rf_use: bool = False) -> None:
...
+            self.block_durations[block_counter] = 0
             if delay_ind_temp[block_counter] > 0:
-                block.delay = SimpleNamespace()
-                block.delay.type = "delay"
-                block.delay.delay = temp_delay_library.data[
-                    delay_ind_temp[block_counter]
-                ]
+                self.block_durations[block_counter] = temp_delay_library.data[
+                        delay_ind_temp[block_counter]
+                    ][0]
```
schuenke, line 260: is this needed?

schuenke (approved):
thanks for the fix @FrankZijlstra 

I reviewed the code and tested it with several old sequences and it seems to work as expected. 

However, IMO it should also work without the lines 257-260, no ?! We override this temp value anyway and it shouldn't matter if we overwrite a temporary delay value or just 0, does it?

In general, the types in event_lib are not all correct and the employed update() method doesn't require `old data`, which you just consequently set to `None` here , but this is something we can discuss during the code refactoring process and not here I guess.

pypulseq/Sequence/read_seq.py
FrankZijlstra: Yes, it is needed. In the old format, delay events were explicitly defined in the event table. In the new format, only the block duration is stored in the event table. However, to calculate the duration, we need the maximum of the stored delay and the actual duration of the block events (determined with `calc_duration`). `get_block` gives us the stored block duration (where you previously got an error), i.e. the delay, which is used by `calc_duration` (but the output of `calc_duration` might be larger depending on the events).

I do feel the whole handling of the delay event is a bit messy in the new format. We define a delay event with `make_delay`, and can use it in `calc_duration`. But as soon as we add it to the sequence, the delay event disappears. Instead, `get_block` will just give us a `block_duration` field, which is a special case in `calc_duration` (`isinstance(event, (float, int))`). As a consequence, the only reason `calc_duration` can determine the duration in this case, is because the duration was already stored.

FrankZijlstra:
> In general, the types in event_lib are not all correct and the employed update() method doesn't require old data, which you just consequently set to None here , but this is something we can discuss during the code refactoring process and not here I guess.
When I removed the use of `old_data` before, I did consider just removing it altogether, but figured that it would be a good idea to keep compatibility (in the unlikely case someone was ever adventurous enough to mess with the event libraries...). But I think it's pretty safe to remove altogether.

pypulseq/Sequence/read_seq.py
schuenke: Ah, thanks for the nice explanation. Wasn't aware that the delay event is gone after adding it to a block.

PR #165, opened 2024-01-18
A few small bugfixes

FrankZijlstra:
- Fixed some rounding errors in `Sequence`
- Fixed possible warning for invalid escape sequence (e.g. `"(\s)+"` -> `r"(\s)+"`)
- Fixed reading back in string definitions that contain spaces (e.g. `seq.set_definition("a", "a a")`)
- Fixed an error in `calculate_kspace` when there are gradients specified on a gradient axis, but there are no ramps (i.e. all gradients are 0)

Edit, two more small bugs:
- Fixed an error with `next_free_block_ID` calculation when reading an empty sequence
- Fixed a bug in `traj_to_grad` and allow it to be used with a multidimensional k-space array

### DIFF
--- a/pypulseq/Sequence/block.py
+++ b/pypulseq/Sequence/block.py
@@ -146,6 +146,9 @@ def set_block(self, block_index: int, *args: SimpleNamespace) -> None:
                     "ref": label_id,
                 }
                 extensions.append(ext)
+        else:
+            # Floating point number given as delay
+            duration = max(duration, event)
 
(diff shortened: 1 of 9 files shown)


FrankZijlstra:
Added a small fix for #166

FrankZijlstra:
And one more: Fixed a bug in `make_trapezoid` that gave an error with a negative `flat_area` (introduced in #146)

PR #170, opened 2024-03-01
Added gradient spectrum calculations

FrankZijlstra:
This PR adds the gradient spectrum calculation as available in MATLAB Pulseq: https://github.com/pulseq/pulseq/blob/master/matlab/demoUnsorted/gradSpectrum.m

The implementation provides a fair number of parameters to control how the calculation is actually done (as we're not entirely sure what is the best way to go about this). The defaults should be sufficient to avoid peaks in the spectrum at the acoustic resonances of your scanner (and avoid damaging the gradient coils).

Basic usage (will plot the spectrums and also return them):
```
seq.calculate_gradient_spectrum()
```

Or including plotting the acoustic resonance ranges from a Siemens .asc file:
```
from pypulseq.utils.siemens.readasc import readasc
from pypulseq.utils.siemens.asc_to_hw import asc_to_acoustic_resonances

asc = readasc(asc_name)
seq.calculate_gradient_spectrum(acoustic_resonances=asc_to_acoustic_resonances(asc[0]))
```

### DIFF
--- a/pypulseq/Sequence/calc_grad_spectrum.py
+++ b/pypulseq/Sequence/calc_grad_spectrum.py
@@ -1,4 +1,4 @@
-from typing import Tuple, List
+from typing import Tuple, List, Union
 
 import numpy as np
 from scipy.signal import spectrogram
@@ -10,7 +10,7 @@ def calculate_gradient_spectrum(
         max_frequency: float = 2000,
         window_width: float = 0.05,
         frequency_oversampling: float = 3,
-        time_range: List[float] | None = None,
+        time_range: Union[List[float], None] = None,
         plot: bool = True,
         combine_mode: str = 'max',
         use_derivative: bool = False,
@@ -94,7 +94,7 @@ def calculate_gradient_spectrum(
     
     # Calculate spectrogram for each gradient channel
     spectrograms: List[np.ndarray] = []
-    spectrogram_rss = 0
+    spectrogram_rss = np.array(0)
     
     for i in range(ng):
         # Use scipy to calculate the spectrograms
--- a/pypulseq/Sequence/calc_grad_spectrum.py
+++ b/pypulseq/Sequence/calc_grad_spectrum.py
@@ -10,7 +10,7 @@ def calculate_gradient_spectrum(
         max_frequency: float = 2000,
         window_width: float = 0.05,
         frequency_oversampling: float = 3,
-        time_range: List[float] = None,
+        time_range: List[float] | None = None,
         plot: bool = True,
         combine_mode: str = 'max',
         use_derivative: bool = False,
@@ -43,7 +43,7 @@ def calculate_gradient_spectrum(
         Whether to plot the spectograms. The default is True.
     combine_mode : str, optional
         How to combine all windows into one spectrogram, options:
-            'max', 'mean', 'sos' (root-sum-of-squares), 'none' (no combination)
+            'max', 'mean', 'rss' (root-sum-of-squares), 'none' (no combination)
         The default is 'max'.
     use_derivative : bool, optional
         Whether the use the derivative of the gradient waveforms instead of the
@@ -57,7 +57,7 @@ def calculate_gradient_spectrum(
     -------
     spectrograms : List[np.ndarray]
         List of spectrograms per gradient channel.
-    spectrogram_sos : np.ndarray
+    spectrogram_rss : np.ndarray
         Root-sum-of-squares combined spectrogram over all gradient channels.
     frequencies : np.ndarray
         Frequency axis of the spectrograms.
@@ -72,7 +72,7 @@ def calculate_gradient_spectrum(
     # Get gradients as piecewise-polynomials
     gw_pp = obj.get_gradients(time_range=time_range)
     ng = len(gw_pp)
-    max_t = max(g.x[-1] for g in gw_pp if g != None)
+    max_t = max(g.x[-1] for g in gw_pp if g is not None)
     
     # Determine sampling points
     if time_range == None:
@@ -93,13 +93,12 @@ def calculate_gradient_spectrum(
         gw = np.diff(gw, axis=1)
     
     # Calculate spectrogram for each gradient channel
-    spectrograms = []
-    frequencies = None
-    spectrogram_sos = 0
+    spectrograms: List[np.ndarray] = []
+    spectrogram_rss = 0
     
     for i in range(ng):
         # Use scipy to calculate the spectrograms
-        freq, times, Sxx = spectrogram(gw[i],
+        freq, times, sxx = spectrogram(gw[i],
                                        fs=1/dt,
                                        mode='magnitude',
                                        nperseg=nwin,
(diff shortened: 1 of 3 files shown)


pypulseq/Sequence/calc_grad_spectrum.py
```diff
@@ -0,0 +1,180 @@
...
+from scipy.signal import spectrogram
+from matplotlib import pyplot as plt
+
+
+def calculate_gradient_spectrum(
+        obj,
+        max_frequency: float = 2000,
+        window_width: float = 0.05,
+        frequency_oversampling: float = 3,
+        time_range: List[float] = None,
```
schuenke, line 13: should be `List[float] | None` or `Union[List[float], None]` if it defauls to None

pypulseq/Sequence/calc_grad_spectrum.py
```diff
@@ -0,0 +1,180 @@
...
+        gw = np.diff(gw, axis=1)
+    
+    # Calculate spectrogram for each gradient channel
+    spectrograms = []
+    frequencies = None
+    spectrogram_sos = 0
+    
+    for i in range(ng):
+        # Use scipy to calculate the spectrograms
+        freq, times, Sxx = spectrogram(gw[i],
```
schuenke, line 102: I know `Sxx` is the official name from scipy.signal.spectrogram, but I would prefer an all lowercase variable name

pypulseq/Sequence/calc_grad_spectrum.py
```diff
@@ -0,0 +1,180 @@
...
+        elif combine_mode == 'none':
+            s = Sxx[mask]
+        else:
+            raise ValueError(f'Unknown value for combine_mode: {combine_mode}, must be one of [max, mean, sos, none]')
+        
+        frequencies = freq[mask]
+        spectrograms.append(s)
+    
+    # Root-sum-of-squares combined spectrogram for all gradient channels
+    spectrogram_sos = np.sqrt(spectrogram_sos)
```
schuenke, line 131: maybe rename to `spectrogram_rss` after applying np.sqrt()

pypulseq/Sequence/calc_grad_spectrum.py
```diff
@@ -0,0 +1,180 @@
...
+    if combine_mode == 'max':
+        spectrogram_sos = spectrogram_sos.max(axis=1)
+    elif combine_mode == 'mean':
+        spectrogram_sos = spectrogram_sos.mean(axis=1)
+    elif combine_mode == 'sos':
+        spectrogram_sos = np.sqrt((spectrogram_sos**2).sum(axis=1))
+    
+    # Plot spectrograms and acoustic resonances if specified
+    if plot:
+        if combine_mode != 'none':
```
schuenke, line 141: `if combine_mode is not None` if we switch to Literal for combine_mode

pypulseq/Sequence/calc_grad_spectrum.py
```diff
@@ -0,0 +1,180 @@
...
+    for i in range(ng):
+        if gw_pp[i] != None:
+            gw[i] = gw_pp[i](t)
+    
+    if use_derivative:
+        gw = np.diff(gw, axis=1)
+    
+    # Calculate spectrogram for each gradient channel
+    spectrograms = []
+    frequencies = None
```
schuenke, line 97: why do you initialize frequencies as `None` here? It should always be a np.ndarray, no?

pypulseq/Sequence/calc_grad_spectrum.py
```diff
@@ -0,0 +1,180 @@
...
+        elif combine_mode == 'mean':
+            s = Sxx[mask].mean(axis=1)
+        elif combine_mode == 'sos':
+            s = np.sqrt((Sxx[mask]**2).sum(axis=1))
+        elif combine_mode == 'none':
+            s = Sxx[mask]
+        else:
+            raise ValueError(f'Unknown value for combine_mode: {combine_mode}, must be one of [max, mean, sos, none]')
+        
+        frequencies = freq[mask]
```
schuenke, line 127: overwriting `frequencies `in every iteration is on purpose?

pypulseq/Sequence/calc_grad_spectrum.py
```diff
@@ -0,0 +1,180 @@
...
+        if gw_pp[i] != None:
+            gw[i] = gw_pp[i](t)
+    
+    if use_derivative:
+        gw = np.diff(gw, axis=1)
+    
+    # Calculate spectrogram for each gradient channel
+    spectrograms = []
+    frequencies = None
+    spectrogram_sos = 0
```
schuenke, line 98: this should also be initialized as np.ndarray

pypulseq/Sequence/calc_grad_spectrum.py
```diff
@@ -0,0 +1,180 @@
...
+    gw = np.zeros((ng,t.shape[0]))
+    for i in range(ng):
+        if gw_pp[i] != None:
+            gw[i] = gw_pp[i](t)
+    
+    if use_derivative:
+        gw = np.diff(gw, axis=1)
+    
+    # Calculate spectrogram for each gradient channel
+    spectrograms = []
```
schuenke, line 96: type-hinting spectrograms as List[np.ndarray] might be helpful

pypulseq/Sequence/calc_grad_spectrum.py
FrankZijlstra: Done

pypulseq/Sequence/calc_grad_spectrum.py
FrankZijlstra: Done

pypulseq/Sequence/calc_grad_spectrum.py
```diff
@@ -0,0 +1,180 @@
...
+
+    """
+    dt = obj.system.grad_raster_time # time raster
+    nwin = round(window_width / dt)
+    nfft = round(frequency_oversampling*nwin)
+
+    # Get gradients as piecewise-polynomials
+    gw_pp = obj.get_gradients(time_range=time_range)
+    ng = len(gw_pp)
+    max_t = max(g.x[-1] for g in gw_pp if g != None)
```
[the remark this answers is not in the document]
FrankZijlstra: Done

pypulseq/Sequence/calc_grad_spectrum.py
FrankZijlstra: Renamed all the `spectrogram_rss` (and renamed 'sos' to 'rss' for consistency). Although the temporary variable technique is sum-of-squares, it would just waste memory to name it differently from the final variable.

pypulseq/Sequence/calc_grad_spectrum.py
FrankZijlstra: Removed the initialization, it is not needed.

pypulseq/Sequence/calc_grad_spectrum.py
FrankZijlstra: Yes, the exact frequencies are not known until `spectrogram` is called, but the result is always the same regardless of the gradient-axis. The same goes for the `times` return value.

pypulseq/Sequence/calc_grad_spectrum.py
FrankZijlstra: Done, but as the shape of the array is unknown until `spectrogram` is called, it can only be initialized to `np.array(0)`.

pypulseq/Sequence/calc_grad_spectrum.py
FrankZijlstra: Done.

FrankZijlstra:
Python output `seq.calculate_gradient_spectrum(max_frequency=10000, combine_mode='rss')` (parameters to be compatible with MATLAB defaults):
image

MATLAB output:
image

pypulseq/Sequence/calc_grad_spectrum.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def calculate_gradient_spectrum(
+        obj,
+        max_frequency: float = 2000,
+        window_width: float = 0.05,
+        frequency_oversampling: float = 3,
+        time_range: List[float] = None,
+        plot: bool = True,
+        combine_mode: str = 'max',
```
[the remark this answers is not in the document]
schuenke: we could still use `combine_mode: Literal["max", "mean", "rss", "none"] = "max"` for better IDE tooltips, but I think we use it nowhere so far. So it's fine for me.

schuenke (approved):

PR #171, opened 2024-03-26
Fix make_extended_trapezoid area

schuenke:
Improve timing/area calculation by ignoring raster_time at first and taking care of it in final calculation. This avoids issues by the discontinuities causing negative flat_times.

The issue accured for example when trying to use `make_extended_trapezoid_area` to calculate the rewinder gradients for spiral MRI sequences.

### DIFF
--- a/pypulseq/make_extended_trapezoid_area.py
+++ b/pypulseq/make_extended_trapezoid_area.py
@@ -76,23 +76,23 @@ def make_extended_trapezoid_area(
         ramp_time_pos_start = round(_calc_ramp_time(max_grad, grad_start) / raster_time)
         ramp_time_neg_start = round(_calc_ramp_time(-max_grad, grad_start) / raster_time)
         max_ramp_time_start = max(ramp_time_pos_start, ramp_time_neg_start)
-        min_time_start = min(max_ramp_time_start, duration - 1)
-        time_ramp_up = np.arange(1, min_time_start + 1)
+        max_time_start = min(max_ramp_time_start, duration - 1)
+        time_ramp_up = np.arange(1, max_time_start + 1)
 
         # Calculate possible ramp-down times for given duration
         ramp_time_pos_end = round(_calc_ramp_time(max_grad, grad_end) / raster_time)
         ramp_time_neg_end = round(_calc_ramp_time(-max_grad, grad_end) / raster_time)
         max_ramp_time_end = max(ramp_time_pos_end, ramp_time_neg_end)
-        min_time_end = min(max_ramp_time_end, duration - 1)
-        time_ramp_down = np.arange(1, min_time_end + 1)
+        max_time_end = min(max_ramp_time_end, duration - 1)
+        time_ramp_down = np.arange(1, max_time_end + 1)
 
         # Create meshgrid of possible ramp-up and ramp-down times
         time_ramp_up, time_ramp_down = np.meshgrid(time_ramp_up, time_ramp_down)
 
-        # calculate corresponding flat times
+        # Calculate corresponding flat times
         flat_time = duration - time_ramp_up - time_ramp_down
 
-        # filter search space for valid timings (flat time >= 0)
+        # Filter search space for valid timings (flat time >= 0)
         valid_indices = flat_time >= 0
         time_ramp_up = time_ramp_up[valid_indices]
         time_ramp_down = time_ramp_down[valid_indices]
@@ -128,27 +128,25 @@ def make_extended_trapezoid_area(
     # binary search below fails to find the optimum in those cases.
     # TODO: Check if range is sufficient, try to calculate the dead space.
     min_duration = max(round(_calc_ramp_time(grad_end, grad_start) / raster_time), 2)
-    max_duration = (
-        max(
-            round(_calc_ramp_time(max_grad, grad_start) / raster_time),
-            round(_calc_ramp_time(-max_grad, grad_start) / raster_time),
-            round(_calc_ramp_time(max_grad, grad_end) / raster_time),
-            round(_calc_ramp_time(-max_grad, grad_end) / raster_time),
-            min_duration,
-        )
-        + 1
-    )
 
+    # Calculate duration needed to ramp down gradient to zero.
+    # From this point onwards, solutions can always be found by extending
+    # the duration and doing a binary search.
+    max_duration = max(
+                       round(_calc_ramp_time(0, grad_start) / raster_time),
+                       round(_calc_ramp_time(0, grad_end) / raster_time),
+                       min_duration,
+                   )
+
+    # Linear search
     solution = None
-    for duration in range(min_duration, max_duration):
+    for duration in range(min_duration, max_duration + 1):
         solution = _find_solution(duration)
         if solution:
             break
 
-    # Perform a binary search if no solution was found
+    # Perform a binary search for duration > max_duration if no solution was found
     if not solution:
-        max_duration = duration
-
         # First, find the upper limit on duration where a solution exists by
         # exponentially expanding the duration.
         while not solution:
--- a/pypulseq/make_extended_trapezoid_area.py
+++ b/pypulseq/make_extended_trapezoid_area.py
@@ -1,6 +1,6 @@
 import math
 from types import SimpleNamespace
-from typing import Tuple
+from typing import Tuple, Union
 
 import numpy as np
 from scipy.optimize import minimize
@@ -11,9 +11,13 @@ from pypulseq.utils.cumsum import cumsum
 
 
 def make_extended_trapezoid_area(
-    area: float, channel: str, grad_end: float, grad_start: float, system: Opts
-) -> Tuple[SimpleNamespace, np.array, np.array]:
-    """Makes the shortest possible extended trapezoid for given area and gradient start and end point.
+    area: float,
+    channel: str,
+    grad_start: float,
+    grad_end: float,
+    system: Union[Opts, None] = None,
+) -> Tuple[SimpleNamespace, np.ndarray, np.ndarray]:
+    """Make (shortest possible) extended trapezoid for given area and gradient start and end point.
 
     Parameters
     ----------
@@ -25,7 +29,7 @@ def make_extended_trapezoid_area(
         Starting non-zero gradient value.
     grad_end : float
         Ending non-zero gradient value.
-    system: Opts
+    system: Opts, optional
         System limits.
 
     Returns
@@ -37,6 +41,9 @@ def make_extended_trapezoid_area(
     amplitude : numpy.ndarray
         Amplitude values of the extended trapezoid.
     """
+    if not system:
+        system = Opts()
+
     slew_rate = system.max_slew * 0.99
     max_grad = system.max_grad * 0.99
     raster_time = system.grad_raster_time
@@ -45,8 +52,8 @@ def make_extended_trapezoid_area(
         return np.ceil(time / raster_time) * raster_time
 
     def _calc_area_and_times(grad_amp, flat_time, slew_rate, grad_start, grad_end):
-        time_ramp_up = _calc_ramp_up_time(grad_amp, slew_rate, grad_start)
-        time_ramp_down = _calc_ramp_down_time(grad_amp, slew_rate, grad_end)
+        time_ramp_up = _calc_ramp_time(grad_amp, grad_start, slew_rate)
+        time_ramp_down = _calc_ramp_time(grad_amp, grad_end, slew_rate)
         area = _calc_area(grad_amp, time_ramp_up, flat_time, time_ramp_down, grad_start, grad_end)
         return area
 
@@ -57,17 +64,17 @@ def make_extended_trapezoid_area(
             + 0.5 * (grad_amp + grad_end) * time_ramp_down
         )
 
-    def _calc_ramp_up_time(grad_amp, slew_rate, grad_start):
-        return abs(grad_amp - grad_start) / slew_rate
-
-    def _calc_ramp_down_time(grad_amp, slew_rate, grad_end):
-        return abs(grad_amp - grad_end) / slew_rate
+    def _calc_ramp_time(grad_1, grad_2, slew_rate):
+        return abs(grad_1 - grad_2) / slew_rate
 
     # first try to find grad_amp without a flat time.
     # for better convergence, we ignore the raster effect here and use a very small time step
     # we will take care of raster effect later on
     flat_time = 0
-    obj1 = lambda x: (area - _calc_area_and_times(x, flat_time, slew_rate, grad_start, grad_end)) ** 2
+
+    def obj1(x):
+        return (area - _calc_area_and_times(x, flat_time, slew_rate, grad_start, grad_end)) ** 2
+
     # try different starting points for the optimization
     optimization_results = [
         minimize(fun=obj1, x0=-max_grad, method="Nelder-Mead"),
@@ -87,7 +94,10 @@ def make_extended_trapezoid_area(
     if obj1_value**0.5 > 1e-3 or abs(grad_amp) > max_grad:
         # Optimization without flat_time did not converge. Try to find flat_time.
         grad_amp = math.copysign(max_grad, grad_amp)
-        obj2 = lambda x: (area - _calc_area_and_times(grad_amp, x, slew_rate, grad_start, grad_end)) ** 2
+
+        def obj2(x):
+            return (area - _calc_area_and_times(grad_amp, x, slew_rate, grad_start, grad_end)) ** 2
+
         res2 = minimize(fun=obj2, x0=system.grad_raster_time, method="Nelder-Mead")
         flat_time = res2.x[0]
         # assert flat_time is larger than 0 and that the objective function is small
@@ -95,17 +105,27 @@ def make_extended_trapezoid_area(
 
     # ensure times are on raster
     flat_time = _to_raster(flat_time)
-    time_ramp_up = _to_raster(_calc_ramp_up_time(grad_amp, slew_rate, grad_start))
-    time_ramp_down = _to_raster(_calc_ramp_down_time(grad_amp, slew_rate, grad_end))
+    time_ramp_up = _to_raster(_calc_ramp_time(grad_amp, grad_start, slew_rate))
+    time_ramp_down = _to_raster(_calc_ramp_time(grad_amp, grad_end, slew_rate))
 
     # recalculate grad_amp while taking raster effect into account
-    obj3 = lambda x: (area - _calc_area(x, time_ramp_up, flat_time, time_ramp_down, grad_start, grad_end)) ** 2
+    def obj3(x):
+        return (area - _calc_area(x, time_ramp_up, flat_time, time_ramp_down, grad_start, grad_end)) ** 2
+
     res = minimize(fun=obj3, x0=grad_amp, method="Nelder-Mead")
     grad_amp = res.x[0]
 
     if flat_time > 0:
-        times = cumsum(0, time_ramp_up, flat_time, time_ramp_down)
-        amplitudes = [grad_start, grad_amp, grad_amp, grad_end]
+        if time_ramp_down > 0:
+            if time_ramp_up > 0:
+                times = cumsum(0, time_ramp_up, flat_time, time_ramp_down)
+                amplitudes = [grad_start, grad_amp, grad_amp, grad_end]
+            else:
+                times = cumsum(0, flat_time, time_ramp_down)
+                amplitudes = [grad_amp, grad_amp, grad_end]
+        else:
+            times = cumsum(0, time_ramp_up, flat_time)
+            amplitudes = [grad_start, grad_amp, grad_end]
     else:
         if time_ramp_down > 0:
             if time_ramp_up > 0:
--- a/pypulseq/make_extended_trapezoid_area.py
+++ b/pypulseq/make_extended_trapezoid_area.py
@@ -13,7 +13,7 @@ def make_extended_trapezoid_area(
     channel: str,
     grad_start: float,
     grad_end: float,
-    system: Union[Opts, None] = None
+    system: Union[Opts, None] = None,
 ) -> Tuple[SimpleNamespace, np.array, np.array]:
     """Make the shortest possible extended trapezoid for given area and gradient start and end point.
 
@@ -41,7 +41,7 @@ def make_extended_trapezoid_area(
     """
     if not system:
         system = Opts()
-        
+
     max_slew = system.max_slew * 0.99
     max_grad = system.max_grad * 0.99
     raster_time = system.grad_raster_time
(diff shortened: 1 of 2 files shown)


FrankZijlstra:
Looks good in principle, works better than the original function in all cases I've tested.

In the code as-is, there are edge-cases in the form of `area == grad_end * raster_time * N` (assuming `grad_start == 0`). That is, where the requested area can be achieved with a triangle gradient between start and end.

In some cases they give an assertion error on (very small) negative `flat_time`:
```pp.make_extended_trapezoid_area(1, 'x', 100000, 0, system)```

In other cases they result in a non-time-optimal solution, due to the rounding to raster times:
```pp.make_extended_trapezoid_area(0.45, 'x', 90000, 0, system)```
Where the optimal solution is given by:
```pp.make_extended_trapezoid('x', times=[0, 1e-5], amplitudes=[0,90000])```

Another set of edge cases is in the form of `area == 0.5 * grad_end**2 / (system.max_slew * 0.99)`  (assuming `grad_start == 0`), where the optimization fails. This is the case where the requested area is equal to the area of ramping up to `grad_end` with 0.99 times the maximum slew rate (as used in the function).
```pp.make_extended_trapezoid_area(0.5 * 100000**2 / (system.max_slew*0.99), 'x', 100000, 0, system)```

An even more rare set of edge cases exist where the taking `ceil` of the ramp up/down and flat times is not optimal. I think the worst case here is that the solution ends up being one or two raster times off the optimum, which is not a big disaster. But the code could in principle consider all 8 solutions (rounding up or down for each of the times).


Besides these edge-cases, an issue I see is that with these changes that remove the discontinuity from the optimization problems, all optimization problems now have closed-form solutions, making all the iterative optimization unnecessary, solutions below:

The first set of optimization:
```
    r = slew_rate
    s = grad_start
    z = grad_end
    v = area
       
    if (s**2 + 2*r*v + z**2) < 0:
        x1 = np.sqrt(-(s**2 + 2*r*v + z**2)) / np.sqrt(2)
    else:
        x1 = np.sqrt(s**2 + 2*r*v + z**2) / np.sqrt(2)
    
    if (s**2 - 2*r*v + z**2) < 0:
        x2 = np.sqrt(-(s**2 - 2*r*v + z**2)) / np.sqrt(2)
    else:
        x2 = np.sqrt(s**2 - 2*r*v + z**2) / np.sqrt(2)
    
    grad_amp_results = [x1, -x1, x2, -x2]
    obj1val_results = [obj1(x) for x in grad_amp_results]
    print(grad_amp_results, obj1val_results)
```
Based on https://www.wolframalpha.com/input?i=%28s-x%29%2Fr*+%28s%2Bx%29%2F2+%2B+%28z-x%29%2Fr*+%28x%2Bz%29%2F2+%3D+v+solve+for+x and https://www.wolframalpha.com/input?i=%28x-s%29%2Fr*+%28s%2Bx%29%2F2+%2B+%28x-z%29%2Fr*+%28x%2Bz%29%2F2+%3D+v+solve+for+x
I think here it should be possible to reduce the solutions to two, or maybe even one, solution, based on whether the requested area is larger or smaller than the area of a triangle gradient between `grad_start` and `grad_end`. E.g. if the area is larger, then `x` must be larger than `max(grad_start,grad_end)`.

The second optimization:
```
        x = grad_amp
        flat_time = -(-2*r*v + s**2 - 2*x**2 + z**2)/(2*r*x)
        print(flat_time)
```
Based on https://www.wolframalpha.com/input?i=%28s-x%29%2Fr*+%28s%2Bx%29%2F2+%2B+f*x+%2B+%28z-x%29%2Fr*+%28x%2Bz%29%2F2+%3D+v+solve+for+f

And the final optimization:
```
    a = time_ramp_up
    f = flat_time
    b = time_ramp_down
    grad_amp = -(a*s + b*z - 2*v)/(a+b+2*f)
```
Based on https://www.wolframalpha.com/input?i=solve+a*%28s%2Bx%29%2F2+%2B+f*x+%2B+b*%28x%2Bz%29%2F2+%3D+v+for+x

These can be easily replaced into the code, obviously with some renaming of the variables I used to make wolfram alpha happy :). These may also solve some of the edge-cases, or make them easier to detect.

File: pypulseq/make_extended_trapezoid_area.py [hunk 1]
```diff
@@ -13,20 +13,18 @@
 def make_extended_trapezoid_area(
     area: float, channel: str, grad_end: float, grad_start: float, system: Opts
```
FrankZijlstra, line 14: While we're at it, change system to be optional with default `system = None`, and initialize with `system = Opts()` below (see other functions for an example).

pypulseq/make_extended_trapezoid_area.py
```diff
@@ -35,100 +33,93 @@ def make_extended_trapezoid_area(
...
-        Tru = (
-            math.ceil(abs(Gp - grad_start) / SR / system.grad_raster_time)
-            * system.grad_raster_time
-        )
-        Trd = (
-            math.ceil(abs(Gp - grad_end) / SR / system.grad_raster_time)
-            * system.grad_raster_time
-        )
-        obj3 = lambda x: (area - __testGA1(x, Tru, Tp, Trd, grad_start, grad_end)) ** 2
+    def _calc_ramp_up_time(grad_amp, slew_rate, grad_start):
```
FrankZijlstra, line 60: `_calc_ramp_up_time` and `_calc_ramp_down_time` are the same function, can just define one `_calc_ramp_time`.

pypulseq/make_extended_trapezoid_area.py
```diff
@@ -35,100 +33,93 @@ def make_extended_trapezoid_area(
...
+                times = cumsum(0, time_ramp_up, time_ramp_down)
+                amplitudes = np.array([grad_start, grad_amp, grad_end])
             else:
-                times = cumsum(0, Trd)
+                times = cumsum(0, time_ramp_down)
                 amplitudes = np.array([grad_start, grad_end])
         else:
-            times = cumsum(0, Tru)
+            times = cumsum(0, time_ramp_up)
             amplitudes = np.array([grad_start, grad_end])
```
FrankZijlstra, line 119: - The `flat_time > 0` section needs checks for `time_ramp_up` and/or `time_ramp_down` being equal to 0.
- The checks may need to consider rounding errors?

Related error:
```pp.make_extended_trapezoid_area(1000, 'x', system.max_grad*0.99, 0, system)```

**Review by FrankZijlstra — CHANGES_REQUESTED** [review]

**Comment by schuenke** [comment]
> Looks good in principle, works better than the original function in all cases I've tested.
> 
> In the code as-is, there are edge-cases in the form of `area == grad_end * raster_time * N` (assuming `grad_start == 0`). That is, where the requested area can be achieved with a triangle gradient between start and end.
> 
> In some cases they give an assertion error on (very small) negative `flat_time`: `pp.make_extended_trapezoid_area(1, 'x', 100000, 0, system)`
> 
> In other cases they result in a non-time-optimal solution, due to the rounding to raster times: `pp.make_extended_trapezoid_area(0.45, 'x', 90000, 0, system)` Where the optimal solution is given by: `pp.make_extended_trapezoid('x', times=[0, 1e-5], amplitudes=[0,90000])`
> 
> Another set of edge cases is in the form of `area == 0.5 * grad_end**2 / (system.max_slew * 0.99)` (assuming `grad_start == 0`), where the optimization fails. This is the case where the requested area is equal to the area of ramping up to `grad_end` with 0.99 times the maximum slew rate (as used in the function). `pp.make_extended_trapezoid_area(0.5 * 100000**2 / (system.max_slew*0.99), 'x', 100000, 0, system)`
>
Yeah, the simple ramp-up or ramp-down cases were and are not handled separately atm. We only differentiate between with or without flat time. Probably this is something we should do directly in the beginning. It should fix all cases mentioned above, no ?! 
> 
> An even more rare set of edge cases exist where the taking `ceil` of the ramp up/down and flat times is not optimal. I think the worst case here is that the solution ends up being one or two raster times off the optimum, which is not a big disaster. But the code could in principle consider all 8 solutions (rounding up or down for each of the times).
>
Rounding down the ramp-times should not be possible, should it? Maybe a recalculation of the flat_time due to larger ramping areas makes sense, but I don't think this will ever lead to deviations of more than one raster_time from the optimum. 
> 
> Besides these edge-cases, an issue I see is that with these changes that remove the discontinuity from the optimization problems, all optimization problems now have closed-form solutions, making all the iterative optimization unnecessary, solutions below:
>
Yeah, we were thinking about the closed-form solution as well, but catching all edge cases, especially when enforcing the raster, is also not super simple and clean. Thus, we decided to provide a simple fix for the current version and maybe implement an alternative version later on. Would be nice to consider nulling the 1st moment as well as for example in https://onlinelibrary.wiley.com/doi/10.1002/mrm.20489

> 
> The first set of optimization:
> 
> ```
>     r = slew_rate
>     s = grad_start
>     z = grad_end
>     v = area
>        
>     if (s**2 + 2*r*v + z**2) < 0:
>         x1 = np.sqrt(-(s**2 + 2*r*v + z**2)) / np.sqrt(2)
>     else:
>         x1 = np.sqrt(s**2 + 2*r*v + z**2) / np.sqrt(2)
>     
>     if (s**2 - 2*r*v + z**2) < 0:
>         x2 = np.sqrt(-(s**2 - 2*r*v + z**2)) / np.sqrt(2)
>     else:
>         x2 = np.sqrt(s**2 - 2*r*v + z**2) / np.sqrt(2)
>     
>     grad_amp_results = [x1, -x1, x2, -x2]
>     obj1val_results = [obj1(x) for x in grad_amp_results]
>     print(grad_amp_results, obj1val_results)
> ```
> 
> Based on https://www.wolframalpha.com/input?i=%28s-x%29%2Fr*+%28s%2Bx%29%2F2+%2B+%28z-x%29%2Fr*+%28x%2Bz%29%2F2+%3D+v+solve+for+x and https://www.wolframalpha.com/input?i=%28x-s%29%2Fr*+%28s%2Bx%29%2F2+%2B+%28x-z%29%2Fr*+%28x%2Bz%29%2F2+%3D+v+solve+for+x I think here it should be possible to reduce the solutions to two, or maybe even one, solution, based on whether the requested area is larger or smaller than the area of a triangle gradient between `grad_start` and `grad_end`. E.g. if the area is larger, then `x` must be larger than `max(grad_start,grad_end)`.
> 
> The second optimization:
> 
> ```
>         x = grad_amp
>         flat_time = -(-2*r*v + s**2 - 2*x**2 + z**2)/(2*r*x)
>         print(flat_time)
> ```
> 
> Based on https://www.wolframalpha.com/input?i=%28s-x%29%2Fr*+%28s%2Bx%29%2F2+%2B+f*x+%2B+%28z-x%29%2Fr*+%28x%2Bz%29%2F2+%3D+v+solve+for+f
> 
> And the final optimization:
> 
> ```
>     a = time_ramp_up
>     f = flat_time
>     b = time_ramp_down
>     grad_amp = -(a*s + b*z - 2*v)/(a+b+2*f)
> ```
> 
> Based on https://www.wolframalpha.com/input?i=solve+a*%28s%2Bx%29%2F2+%2B+f*x+%2B+b*%28x%2Bz%29%2F2+%3D+v+for+x
> 
> These can be easily replaced into the code, obviously with some renaming of the variables I used to make wolfram alpha happy :). These may also solve some of the edge-cases, or make them easier to detect.



For the moment, I would prefer to stick with the old style using iterative optimization and just change the other points you mentioned. I will adjust the code and push the changes soon.

**Comment by fzimmermann89** [comment]
> An even more rare set of edge cases exist where the taking `ceil` of the ramp up/down and flat times is not optimal. I think the worst case here is that the solution ends up being one or two raster times off the optimum, which is not a big disaster. But the code could in principle consider all 8 solutions (rounding up or down for each of the times).

The calculated time point is the earliest possible time to achieve the area while satisfying the slope constraint.
Rounding down would always slightly exceed the maximum slope afaik.
Thus, we have to always round up, right?

**Comment by FrankZijlstra** [comment]
> Yeah, the simple ramp-up or ramp-down cases were and are not handled separately atm. We only differentiate between with or without flat time. Probably this is something we should do directly in the beginning. It should fix all cases mentioned above, no ?!
All, except the final case, where the triangle is achievable with unrounded ramp times, but not with rounded ramp times. In the first closed form solution, this is the case where one of the discriminants is exactly zero. That is, any solution `grad_start <= x <= grad_end` would yield an objective of exactly 0.

> Rounding down the ramp-times should not be possible, should it? Maybe a recalculation of the flat_time due to larger ramping areas makes sense, but I don't think this will ever lead to deviations of more than one raster_time from the optimum.
It's not a big issue, I agree that it will be on the order of one or two raster times off the optimum, but it is possible. If you round up any of the times, the gradient area increases, and `grad_amp` will need to come down as a result. This lower amplitude can mean a ramp time that was just above the raster time could come down to just below.

> Yeah, we were thinking about the closed-form solution as well, but catching all edge cases, especially when enforcing the raster, is also not super simple and clean. Thus, we decided to provide a simple fix for the current version and maybe implement an alternative version later on. Would be nice to consider nulling the 1st moment as well as for example in https://onlinelibrary.wiley.com/doi/10.1002/mrm.20489

I don't think the solutions I provided introduce any new edge cases. In fact, I think it makes it easier to detect the edge cases (i.e. avoid divisions by 0).

> For the moment, I would prefer to stick with the old style using iterative optimization and just change the other points you mentioned. I will adjust the code and push the changes soon.

All that needs to happen on the analytical solutions is a little cleanup as I just hacked them together :). But they do work fine!

**Comment by FrankZijlstra** [comment]
Little update on the analytical solutions:

First optimization, simplified and added a check for the case where the gradient is triangular (still needs to be preceded with a check that the solution is not just a triangular gradient with a ramp time that fits raster_time):
```
    if area == (s+z)/2 * abs(s-z)/r:
        # Solution is triangular gradient, set grad_amp to violate max_grad, so flat_time optimization follows
        if area > (s+z)/2 * _to_raster(abs(s-z)/r):
            grad_amp = -max_grad*2
        else:
            grad_amp = max_grad*2
    elif area > (s+z)/2 * abs(s-z)/r:
        grad_amp = np.sqrt(s**2 + 2*r*v + z**2) / np.sqrt(2)
    else:
        grad_amp = -np.sqrt(s**2 - 2*r*v + z**2) / np.sqrt(2)
```

Flat time optimization needed a second case:
```
        x = grad_amp
        if x < 0:
            flat_time = -(-2*r*v + s**2 - 2*x**2 + z**2)/(2*r*x)
        else:
            flat_time = (2*r*v + s**2 - 2*x**2 + z**2)/(2*r*x)
```

In addition, I wrote a little pytest script for the function, with some of the things I tried, I suggest you add it to pypulseq/tests, and add test cases for things I have not covered, or if new things come up:
[test_make_extended_trapezoid_area.zip](https://github.com/imr-framework/pypulseq/files/14873579/test_make_extended_trapezoid_area.zip)

**Comment by FrankZijlstra** [comment]
Bigger update... Because I was curious about the impact of rounding up or down, I added a test function based on recreating an existing extended gradient. It finds many failure cases, some 1e-5 off on the timing, but some a few that are a multiple of the duration off. I'm not sure what exactly these cases represent, but that the come up in a test with randomly generated values is a bit worrying.

[test_make_extended_trapezoid_area.zip](https://github.com/imr-framework/pypulseq/files/14874487/test_make_extended_trapezoid_area.zip)

**Comment by btasdelen** [comment]
@schuenke Just as a heads-up, I don't know how it will perform with the recent changes to `make_extended_trapezoid_area`, but when I tried using this approach for Spiral rewinders, accumulated errors in k-space were large enough to see some artifacts (for bSSFP), which I believe comes from rounding up to the raster time. My guess is there might be another solution that has less error when round-up, but we won't know as the optimization assumes continuous case.

Another issue was the rewinders were significantly longer than they need to be. At the end of the day, I started using [GrOpt](https://github.com/cmr-group/gropt), and I am quite happy with the result.

**Comment by fzimmermann89** [comment]
The goal of the rewrite was also to reduce this error..

Once the different time points are determined, the last step should find the optimal gradient amplitude given the (rounded to raster time) time points to achieve exactly the requested amplitude.

It should also always find the shortest possible gradient up to (worst case) 3 gradient raster steps.

**Comment by btasdelen** [comment]
@fzimmermann89 Correct me if I'm wrong, but it is the shortest **trapezoid**, not the shortest gradient, that nulls the M0.

**Comment by fzimmermann89** [comment]
> @fzimmermann89 Correct me if I'm wrong, but it is the shortest **trapezoid**, not the shortest gradient, that nulls the M0.

Maybe I am wrong, but I'd say it should be the shortest **gradient** that has the requested area.
Ignoring raster_time  for now, the shortest possible should rise with the maximum allowed slew rate. If there was a short gradient that has somewhere a slope smaller than the max_slew rate, you can always make it shorter by rising faster.  So the fastest gradient to null the 0-th moment will be the trapezoid.  See also, for example, https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7606655/figure/F2/.

The only difference can be of the order of few (~3) gradient_raster steps ..

Or where is my error? How would a shorter gradient look like that has the same area as a max-slew trapezoid?

**Comment by btasdelen** [comment]
@fzimmermann89 You might be right, and it might be a problem with my implementation back then that I was not able to find the shortest gradient. I can not think of any case that there is a shorter gradient.

**Comment by schuenke** [comment]
> @schuenke Just as a heads-up, I don't know how it will perform with the recent changes to `make_extended_trapezoid_area`, but when I tried using this approach for Spiral rewinders, accumulated errors in k-space were large enough to see some artifacts (for bSSFP), which I believe comes from rounding up to the raster time. My guess is there might be another solution that has less error when round-up, but we won't know as the optimization assumes continuous case.
> 
> Another issue was the rewinders were significantly longer than they need to be. At the end of the day, I started using [GrOpt](https://github.com/cmr-group/gropt), and I am quite happy with the result.

Thanks for the feedback. I'll keep it in mind, but I would definitely prefer an elegant and efficient solution in pypulseq. At least for the "simple" 0-th moment nulling. Pretty optimistic we can get this done 💪

**Comment by schuenke** [comment]
> ```
>     if area == (s+z)/2 * abs(s-z)/r:
>         # Solution is triangular gradient, set grad_amp to violate max_grad, so flat_time optimization follows
>         if area > (s+z)/2 * _to_raster(abs(s-z)/r):
>             grad_amp = -max_grad*2
>         else:
>             grad_amp = max_grad*2
>     elif area > (s+z)/2 * abs(s-z)/r:
>         grad_amp = np.sqrt(s**2 + 2*r*v + z**2) / np.sqrt(2)
>     else:
>         grad_amp = -np.sqrt(s**2 - 2*r*v + z**2) / np.sqrt(2)
> ```

Hmm, not sure about this. 
Lets define `ramp_time = abs(s-z)/r` as the ramp time, which is NOT on raster

`_to_raster(ramp_time)` will always be larger than `ramp_time`, because we use `np.ceil()`
This means `area > (s+z)/2 * _to_raster(abs(s-z)/r)` should never be reachable, no ?!

IMO we only want to use a single triangular gradient (just ONE ramp) if `area == (s+z)/2 * _to_raster(abs(s-z)/r)`, which is a very rare case. 

Just to make clear we all have the same things in mind... we expect the following cases:
1) a grad with a single ramp from grad_start to grad_end, which could potentially be of length 1 * raster_time
2) a grad with 2 ramps, but no flat time. This means one ramp from grad_start to grad_amp and another from grad_amp to grad_end, with arbitrary signs of grad_start/grad_amp/grad_end. Note: this might be the same as case 1 (for length >= 2 * raster_time), if the slopes of both ramps are identical. 
3) a grad with 2 slopes (ramp_up and ramp_down) and a flat time
4) a grad with ramp_up and a flat_time, but no ramp_down
5) a grad with a flat_time and ramp_down, but no ramp_up

**Comment by FrankZijlstra** [comment]
> Hmm, not sure about this. Lets define `ramp_time = abs(s-z)/r` as the ramp time, which is NOT on raster
> 
> `_to_raster(ramp_time)` will always be larger than `ramp_time`, because we use `np.ceil()` This means `area > (s+z)/2 * _to_raster(abs(s-z)/r)` should never be reachable, no ?!
> 
> IMO we only want to use a single triangular gradient (just ONE ramp) if `area == (s+z)/2 * _to_raster(abs(s-z)/r)`, which is a very rare case.
`s+z` can be negative. In plain text it might be easier to understand:
If the requested area is equal to the area of a triangle between start and end without rounding, then (assuming we already checked that it wasn't the one special case that you mention) we need to check if the desired area is larger or smaller than the area of a triangle between start and end without rounding, which will inform whether grad_amp should be +grad_max or -grad_max. And then we solve for `flat_time`.

But I just realize there is still a problem with that case, because that does not necessarily mean `flat_time > 0`. It will get fixed later by the third optimization, but the ramp times will be too long.

Just to be clear, this is not a problem with just the analytical solution, the same thing happens when using optimization.

> Just to make clear we all have the same things in mind... we expect the following cases:
> 
>     1. a grad with a single ramp from grad_start to grad_end, which could potentially be of length 1 * raster_time
> 
>     2. a grad with 2 ramps, but no flat time. This means one ramp from grad_start to grad_amp and another from grad_amp to grad_end, with arbitrary signs of grad_start/grad_amp/grad_end. Note: this might be the same as case 1 (for length >= 2 * raster_time), if the slopes of both ramps are identical.
> 
>     3. a grad with 2 slopes (ramp_up and ramp_down) and a flat time
> 
>     4. a grad with ramp_up and a flat_time, but no ramp_down
> 
>     5. a grad with a flat_time and ramp_down, but no ramp_up

On case 1, I assume you mean (integer) N * raster_time. And if you check for case 1, then a situation with equal slopes in case 2 will never happen. And it shouldn't, because the optimization problem is ill-defined in that case, every solution `grad_start <= grad_amp <= grad_end` will be identical. And with that case excluded, the optimization problem is much narrower with only two possibilities, because we already know whether `grad_amp` should be larger or smaller than `max(grad_start, grad_end)`.

And in case 4 and 5, I noticed issues with start or end set to `max_grad*0.99`: the initial `grad_amp` solution can be `max_grad*0.99`, which would mean there is no ramp up or ramp down. But the `flat_time` is not rounded yet, which would change the area, and change `grad_amp`, and therefore require both ramp up and ramp down.

**Comment by schuenke** [comment]
> > Hmm, not sure about this. Lets define `ramp_time = abs(s-z)/r` as the ramp time, which is NOT on raster
> > `_to_raster(ramp_time)` will always be larger than `ramp_time`, because we use `np.ceil()` This means `area > (s+z)/2 * _to_raster(abs(s-z)/r)` should never be reachable, no ?!
> > IMO we only want to use a single triangular gradient (just ONE ramp) if `area == (s+z)/2 * _to_raster(abs(s-z)/r)`, which is a very rare case.
> 
> `s+z` can be negative. In plain text it might be easier to understand: If the requested area is equal to the area of a triangle between start and end without rounding, then (assuming we already checked that it wasn't the one special case that you mention) we need to check if the desired area is larger or smaller than the area of a triangle between start and end with rounding, which will inform whether grad_amp should be +grad_max or -grad_max. And then we solve for `flat_time`.
> 
> But I just realize there is still a problem with that case, because that does not necessarily mean `flat_time > 0`. It will get fixed later by the third optimization, but the ramp times will be too long.
> 
> Just to be clear, this is not a problem with just the analytical solution, the same thing happens when using optimization.
> 
> > Just to make clear we all have the same things in mind... we expect the following cases:
> > ```
> > 1. a grad with a single ramp from grad_start to grad_end, which could potentially be of length 1 * raster_time
> > 
> > 2. a grad with 2 ramps, but no flat time. This means one ramp from grad_start to grad_amp and another from grad_amp to grad_end, with arbitrary signs of grad_start/grad_amp/grad_end. Note: this might be the same as case 1 (for length >= 2 * raster_time), if the slopes of both ramps are identical.
> > 
> > 3. a grad with 2 slopes (ramp_up and ramp_down) and a flat time
> > 
> > 4. a grad with ramp_up and a flat_time, but no ramp_down
> > 
> > 5. a grad with a flat_time and ramp_down, but no ramp_up
> > ```
> 
> On case 1, I assume you mean (integer) N * raster_time. And if you check for case 1, then a situation with equal slopes in case 2 will never happen. And it shouldn't, because the optimization problem is ill-defined in that case, every solution `grad_start <= grad_amp <= grad_end` will be identical. And with that case excluded, the optimization problem is much narrower with only two possibilities, because we already know whether `grad_amp` should be larger or smaller than `max(grad_start, grad_end)`.
> 
> And in case 4 and 5, I noticed issues with start or end set to `max_grad*0.99`: the initial `grad_amp` solution can be `max_grad*0.99`, which would mean there is no ramp up or ramp down. But the `flat_time` is not rounded yet, which would change the area, and change `grad_amp`, and therefore require both ramp up and ramp down.
> 
> To be clear, I'm not expecting every edge case to be solved per se, but at the very least we can mark where there are issues. Right now I'm most worried about my second set of tests (recreating existing gradients), giving more errors than with the original code.

Ah, sorry I somehow thought there was an abs() around the `s+z`... makes sense.

Just to clarify: I really like the idea of having a closed-form solution and I am motivated to implement it. I just want to think of potential edge-cases before the implementation.

Regarding the errors in the test: you also test for existing gradient without max_slew, don't you? There might be cases where using max_slew is not optimal. E.g. if area is slightly larger than the area of the simple single ramp, but using a slightly lower slew rate might do the job?! Or Is there always a solution with max_slew that should result in the same total time?

I will think about this again and try to define some example cases 😄 




```

FrankZijlstra:
> Just to clarify: I really like the idea of having a closed-form solution and I am motivated to implement it. I just want to think of potential edge-cases before the implementation.
> 
> Regarding the errors in the test: you also test for existing gradient without max_slew, don't you? There might be cases where using max_slew is not optimal. E.g. if area is slightly larger than the area of the simple single ramp, but using a slightly lower slew rate might do the job?! Or Is there always a solution with max_slew that should result in the same total time?
> 
> I will think about this again and try to define some example cases 😄

I have also found myself strangely intrigued by this seemingly simple geometric problem ;). I am currently looking into the cases where the total duration is multiple times off the optimum.
E.g.
```g,_,_ = pp.make_extended_trapezoid_area(channel='x', grad_start=-1680000, grad_end=-1000000, area=-100, system=system)```
Which has a duration of 69e-5, while I know the optimum should be around 10e-5...

And no, I have not explicitly considered that not using `max_slew` could be more optimal. If this is the case, then not rounding the ramp times in the initial optimizations may create additional issues.

**Comment by FrankZijlstra** [comment]
So, it appears the issue at hand is that the first optimization problem can have at least 3 solutions which achieve a given area:
image
(y = `area`, x = `grad_amp`)

These will have near-zero, but not identical objective values, so it's essentially random which one gets picked. But they do not have the same duration. In a bad case, the longest duration gets picked, which is far off the optimum.

In addition, the solver does not always find these solutions. I think the tolerances on the optimization should be lower (at the cost of time), and the 0 initialization is always a stationary point, so that needs to be split into small positive and small negative initializations.

This is actually a good reason to use the closed-form solutions. But I think we need to consider all 4 possible solutions (`[x1, -x1, x2, -x2]` like in my first comment), and use the one with the shortest total time.

**Comment by FrankZijlstra** [comment]
In the closed-form solution of the third problem, this fixes any slew rate violations:
```
    a = time_ramp_up
    f = flat_time
    b = time_ramp_down
    grad_amp_new = -(a*s + b*z - 2*v)/(a+b+2*f)
    
    changed = True
    
    while changed:
        time_ramp_up_new = _to_raster(_calc_ramp_up_time(grad_amp_new, slew_rate, grad_start))
        time_ramp_down_new = _to_raster(_calc_ramp_down_time(grad_amp_new, slew_rate, grad_end))
        if time_ramp_up_new > time_ramp_up or time_ramp_down_new > time_ramp_down:
            a = time_ramp_up_new
            f = flat_time
            b = time_ramp_down_new
            time_ramp_up = time_ramp_up_new
            time_ramp_down = time_ramp_down_new
            
            grad_amp_new = -(a*s + b*z - 2*v)/(a+b+2*f)
        else:
            changed = False
    
    grad_amp = grad_amp_new
```

Even when only rounding up, `grad_amp` can increase in some cases. So you need to keep recalculating until the ramp times are no longer increasing.

**Comment by FrankZijlstra** [comment]
@schuenke I pushed the fixed version. I went back to using a grid search to solve for the timing. It's not as efficient as it can be, but for normal system parameters it is fine. For low slew rate (relative to maximum gradient strength), it can be slow to find a solution (but it will).

**Comment by schuenke** [comment]
> @schuenke I pushed the fixed version. I went back to using a grid search to solve for the timing. It's not as efficient as it can be, but for normal system parameters it is fine. For low slew rate (relative to maximum gradient strength), it can be slow to find a solution (but it will).

Nice work 👍 While reviewing the code, I added some type hints, documentation and introduced some temp vars to improve the readability. Feel free to revert changes you don't like.

**Comment by FrankZijlstra** [comment]
Looks good. I tightened the bound for the linear search a bit by only considering a ramp down/up to gradient strength 0. After this, the desired area can always be reached by extending the duration (so binary search works there). Also did some minor cleaning.

I had a look at the dead space issue. Without considering the raster time it is possible to calculate the range where the area is unreachable with some ugly quadratic formula. But the bounds it find are not entirely accurate once the raster time is considered. I do think, that if this can be solved accurately, it gives only one or two possible solutions for the entire problem (i.e. find the points where the minimum or maximum area for a given duration is equal to the desired area). But for now I'll leave it as is, we can revisit it if the performance of the function turns out to be too poor for some situations.

File: pypulseq/make_extended_trapezoid_area.py [hunk 4]
```diff
@@ -41,104 +41,132 @@ def make_extended_trapezoid_area(
...
+        -------
+            Tuple of ramp-up time, flat time, ramp-down time, gradient amplitude or None if no solution was found
+
+        Raises
+        ------
+            ValueError if no solution was found that satisfies the constraints and the desired area.
+        """
+
+        # Calculate possible ramp-up times for given duration
+        ramp_time_pos_start = round(_calc_ramp_time(max_grad, grad_start) / raster_time)
```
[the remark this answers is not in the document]
fzimmermann89: time-> steps ?

pypulseq/make_extended_trapezoid_area.py
```diff
@@ -1,134 +1,188 @@
...
+
+            if fun(test_value):
+                return binary_search(fun, lower_limit, test_value)
+            else:
+                return binary_search(fun, test_value, upper_limit)
 
-        res = minimize(fun=obj3, x0=Gp, method="Nelder-Mead")
-        Gp, obj3val = *res.x, res.fun
-        assert obj3val < 1e-3  # Did the final search converge?
+        solution = binary_search(_find_solution, max_duration // 2, max_duration)
```
fzimmermann89, line 167: ?

pypulseq/make_extended_trapezoid_area.py
```diff
@@ -1,134 +1,188 @@
...
+    # the duration and doing a binary search.
+    max_duration = max(
+                       round(_calc_ramp_time(0, grad_start) / raster_time),
+                       round(_calc_ramp_time(0, grad_end) / raster_time),
+                       min_duration,
+                   )
+
+    # Linear search
+    solution = None
+    for duration in range(min_duration, max_duration + 1):
```
fzimmermann89, line 143: ?

FrankZijlstra:
Made one final change that substantially narrows the timings that are checked by `_find_solution`.

FrankZijlstra (approved):

Issue #169, opened 2024-02-14
"Identical" shapes not found in event_lib


schuenke:
**Describe the bug**
When using RF pulses with identical settings but varying flip angles, pypulseq does not recognize the shapes to be identical and adds new shape IDs for each pulse.  For example in MRF sequences this leads to huge seq-files that cannot be executed at the scanner. 

**To Reproduce**
Create a simple sequence with 2 rf pulses that have the same settings, but flip angles of 0.1 and 90 for example. 

Due to the [calculation ](https://github.com/imr-framework/pypulseq/blob/20cce84c00fa8ffd637685a44746e0910b655edd/pypulseq/Sequence/block.py#L605-L607) of the magnitude of the pulses using:
```python
    mag = np.abs(event.signal)
    amplitude = np.max(mag)
    mag /= amplitude
```
we end up with different `mag` arrays and thus also with different compressed shapes.

As the [find_or_insert](https://github.com/imr-framework/pypulseq/blob/20cce84c00fa8ffd637685a44746e0910b655edd/pypulseq/event_lib.py#L75) method doesn't take any tolerances into account, the shape is not found and a new one is inserted

**Expected behavior**
The shape_id should be independent from the flip angle / precision errors.

**Potential fix**
I would suggest to accept some tolerances and simply change [line 100 in the find_or_insert method](https://github.com/imr-framework/pypulseq/blob/20cce84c00fa8ffd637685a44746e0910b655edd/pypulseq/event_lib.py#L100) from
```python
new_data = np.asarray(new_data)
```
to
```python
new_data = np.round(new_data, 12)
```

For me, rounding it to 12 decimal places was enough to ensure all shapes are found to be the same.
This is why I would suggest to include this directly in the event_lib method and not just in `register_rf_event` etc.

schuenke:
Update: In my opinion it actually makes sense to reduce the precision to 9 decimal places, because this seems to be the precision of the values that are written to the seq-file

btasdelen:
Hi @schuenke, we faced this exact issue before. At the time, we just worked around it by using forcing them to be registered as the same shape with `register_rf_event`.

I think rounding to 9 decimal places is very reasonable. I would also guess how many decimal places would result in identical arrays will depend on the scales of the waveforms. 

Maybe we should also do the rounding in `make_rf` functions to keep the waveforms consistent?

FrankZijlstra:
[...]

PR #178, opened 2024-06-06
Added functionality to remove duplicate events

FrankZijlstra:
This PR provides a basic fix for issue #169. Both `Sequence` and `EventLibrary` now have a `remove_duplicates` function that removes events that would be identical after rounding. The function is called both before `Sequence.write` and after `Sequence.read` to ensure that no actual duplicate events exist in the event libraries.

I opted not to do any rounding while adding events, because this affects all sequence analysis functions (e.g. `calculate_kspace`), and it affects any use of `get_block` to copy events (i.e. there can be compounding rounding errors if events are stored rounded). The proposed solution is also a lot faster.

This does mean that in the example of #169, the library will still contain a lot of events, which will only be filtered when writing the sequence. Alternatively, `seq.remove_duplicates(in_place=True)` could be called manually at any point during the creation of a sequence to force rounding of the event libraries.

In addition, I added a basis for `Sequence` unittests, testing whether sequences can be write and read back in, and recreated by `add_block(get_block(...))`, and still produce (within rounding errors) the same sequence (tested with `get_block`, `get_gradients`, and `calculate_kspace`). It also checks the .seq output compared to a known baseline, contained in the `tests/expected_output` directory. The expected .seq output files can be regenerated with the current repository by setting a SAVE_EXPECTED environment variable before running pytest (e.g. if a change in the output happens that is confirmed to be correct).

These tests should be expanded with further sequences snippets which tests all pypulseq functionality, and e.g. including all current example sequences.

Adding these tests uncovered two small bugs fixes in `Sequence` that I included here.


pypulseq/Sequence/sequence.py
```diff
@@ -1177,6 +1178,76 @@ def register_label_event(self, event: SimpleNamespace) -> int:
...
+            tmp = self.block_cache
+            self.block_cache = {}
+            seq_copy = deepcopy(self)
+            self.block_cache = tmp
+
+        # Find duplicate in shape library
+        seq_copy.shape_library, mapping = seq_copy.shape_library.remove_duplicates(9)
+
+        # Remap shape IDs of arbitrary gradient events
+        for x in seq_copy.grad_library.data:
```
schuenke, line 1211: maybe rename `x` to `grad_event` etc. for better readability

pypulseq/Sequence/sequence.py
```diff
@@ -1177,6 +1178,76 @@ def register_label_event(self, event: SimpleNamespace) -> int:
...
+            if seq_copy.grad_library.type[x] == 'g':
+                data = seq_copy.grad_library.data[x]
+                new_data = (data[0],) + (mapping[data[1]], mapping[data[2]]) + data[3:]
+                if data != new_data:
+                    seq_copy.grad_library.update(x, None, new_data)
+
+        # Remap shape IDs of RF events
+        for x in seq_copy.rf_library.data:
```
schuenke, line 1219: same as above: rename `x` -> `rf_event`

pypulseq/Sequence/sequence.py
```diff
@@ -1177,6 +1178,76 @@ def register_label_event(self, event: SimpleNamespace) -> int:
...
+            data = seq_copy.rf_library.data[x]
+            new_data = (data[0],) + (mapping[data[1]], mapping[data[2]], mapping[data[3]]) + data[4:]
+            if data != new_data:
+                seq_copy.rf_library.update(x, None, new_data)
+
+        # Filter duplicates in gradient library
+        seq_copy.grad_library, mapping = seq_copy.grad_library.remove_duplicates((6, 6, 6, 6, 6, 6))
+
+        # Remap gradient event IDs
+        for x in seq_copy.block_events:
```
schuenke, line 1229: `x` -> `block` (below as well)

pypulseq/event_lib.py
```diff
@@ -244,3 +246,67 @@ def update_data(
...
+            Round the data array to a specified number of significant digits,
+            specified by `digits`. Rounding behaviour is similar to the {.Ng}
+            format specifier if N > 0, and similar to {.0f} otherwise.
+            """
+            mags = 10 ** (digits - (np.ceil(np.log10(abs(data) + 1e-12))) if digits > 0 else 0)
+            result = np.round(data * mags) / mags
+            result.flags.writeable = False
+            return result
+
+        # Round library data bbased on `digits` specification
```
schuenke, line 294: `bbased` -> `based`

schuenke (approved):
I tested the Code with the problematic sequence from #169 and everyhing works fine. I didn't check all tests in detail, but the idea totally makes sense to me. 

My comments are not meant as necessary changes, but only as a suggestion for better readability/debugging.

PR #180, opened 2024-06-17
feat: add TRID label (GE interpreter).

mcencini:
This PR adds "TRID" label (see pulseq/pulseq@682b9f1) to allow the GE interpreter to optimize the sequence by grouping consecutive blocks, removing the dead times between them. This is required to allow accurate execution of sequences designed for Siemens systems on GE systems.


mcencini:
Hi all, I just noticed that there was a "np.NaN" instead of "np.nan" which is now deprecated in the new Numpy 2.0 release (from 2 days ago). I have now fixed this

FrankZijlstra:
Thanks for the NaN fix, was going to do it in a separate PR! Not sure why it's not running the unit tests now, but I don't expect any issue with an additional label.

PR #188, opened 2024-08-08
RF: Move sigpy to an extra optional dependency

wtclarke:
- Vendor our own adiabatic pulse functions in `make_adiabatic_pulse.py`
- Only enable `make_sigpy_pulse.py` if sigpy is installed as an extra.
- Add test function for `make_adiabatic_pulse.py`
- Add markers to disable tests on sigpy pulses if not present
- Update the CI tests.

### DIFF
--- a/pypulseq/make_adiabatic_pulse.py
+++ b/pypulseq/make_adiabatic_pulse.py
@@ -4,7 +4,6 @@ from copy import copy
 
 import numpy as np
 import math
-from sigpy.mri.rf import hypsec, wurst
 
 from pypulseq import eps
 from pypulseq.calc_duration import calc_duration
@@ -136,7 +135,7 @@ def make_adiabatic_pulse(
         If invalid pulse use is encountered.
         If slice thickness is not provided but slice-selective trapezoid event is expected.
     """
-    if system == None:
+    if system is None:
         system = Opts.default
         
     valid_pulse_types = ["hypsec", "wurst"]
@@ -154,15 +153,13 @@ def make_adiabatic_pulse(
         dwell = system.rf_raster_time
 
     n_raw = round(duration / dwell + eps)
-    # Number of points must be divisible by 4 - requirement of sigpy.mri
+    # Number of points must be divisible by 4 - requirement of individual pulse functions
     N = math.floor(n_raw / 4) * 4
 
     if pulse_type == "hypsec":
-        am, fm = hypsec(n=N, beta=beta, mu=mu, dur=duration)
+        am, fm = _hypsec(n=N, beta=beta, mu=mu, dur=duration)
     elif pulse_type == "wurst":
-        am, fm = wurst(n=N, n_fac=n_fac, bw=bandwidth, dur=duration)
-    else:
-        raise ValueError("Unsupported adiabatic pulse type.")
+        am, fm = _wurst(n=N, n_fac=n_fac, bw=bandwidth, dur=duration)
 
     pm = np.cumsum(fm) * dwell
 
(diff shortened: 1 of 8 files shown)


wtclarke:
@schuenke All good for review.

pypulseq/make_adiabatic_pulse.py
schuenke, line 1: I think in general we should default parameters, that are overwritten with system attributes in functions (like raster times, dwell times) as None instead of 0. For `make_adiabatic_pulse()`, this would be `dwell`, `max_grad` and `max_slew` imo.

schuenke (approved):
I tried to refactor the code a little bit by using more descriptive variable names, introducing some variable for calculations etc...

I also realized that the padding code was not working, but I think my refactored version is a bit cleaner / easier to read anyway. 

I added some additional comments. Maybe you can have a look.

pypulseq/make_adiabatic_pulse.py
```diff
@@ -136,7 +135,7 @@ def make_adiabatic_pulse(
         If invalid pulse use is encountered.
         If slice thickness is not provided but slice-selective trapezoid event is expected.
     """
-    if system == None:
+    if system is None:
         system = Opts.default
         
     valid_pulse_types = ["hypsec", "wurst"]
```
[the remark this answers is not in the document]
wtclarke: Yes, `if pulse_type != "" and pulse_type not in valid_pulse_types:` should probably be `if not pulse_type or pulse_type not in valid_pulse_types:`

pypulseq/make_adiabatic_pulse.py
```diff
@@ -154,15 +153,13 @@ def make_adiabatic_pulse(
         dwell = system.rf_raster_time
 
     n_raw = round(duration / dwell + eps)
```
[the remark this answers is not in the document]
wtclarke: No idea. I think we should probably leave it as I presume it was used to avoid some bug, even if it's not the best way of doing so.

pypulseq/make_adiabatic_pulse.py
wtclarke: Yes.

And instead of 
```
        if max_grad > 0:
            system = copy(system)
            system.max_grad = max_grad

        if max_slew > 0:
            system = copy(system)
            system.max_slew = max_slew
```
Where it copies system and updates it, it should just pass these parameters through to `make_trapezoid`.

wtclarke:
Thanks for reviewing this in depth. As a general rule, don't refactor additional code in this sort of PR as it just complicates the history. E.g. If we find a bug down the line, is it a optional variable rename or was it me removing the sigpy dep?

And remember [PEP8 unto thyself, not unto others](https://youtu.be/wf-BqAjZb8M?si=0qUKHOZ291x_2nQl&t=1200) (a video well worth watching).

pypulseq/make_adiabatic_pulse.py
schuenke: sounds good 👍

schuenke (approved):

schuenke:
> Thanks for reviewing this in depth. As a general rule, don't refactor additional code in this sort of PR as it just complicates the history. E.g. If we find a bug down the line, is it a optional variable rename or was it me removing the sigpy dep?
> 
> And remember [PEP8 unto thyself, not unto others](https://youtu.be/wf-BqAjZb8M?si=0qUKHOZ291x_2nQl&t=1200) (a video well worth watching).

Well, in general I agree, but the current PyPulseq code is basically not tested at all and far away from beeing stable. 

Sure, I could have created another PR for the refactoring. And another one for the bug fixe(s). And maybe even a next one for the variable renaming. But in all cases our tests would have passed (although the code was not even working - see the padding code that used invalid numpy function calls) and future debugging wouldn't have been easier imo. 

So for the moment, I would still refactor the code in every file that is touched by a PR anyway. To be honest, I think this will also lead to a better code review, because I can almost guarantee that nobody will ever review a PR that refactors all modules and functions. At least not in detail.

wtclarke:
Fair points. Are you able to merge? Or is it someone else?

PR #189, opened 2024-08-09
Add area attribute for arbitrary gradients

schuenke:
The relevant change is the introduced area attribute:
`grad.area = (waveform * system.grad_raster_time).sum()`

The rest is minor refactoring and documentation.

### DIFF
--- a/pypulseq/make_arbitrary_grad.py
+++ b/pypulseq/make_arbitrary_grad.py
@@ -30,12 +30,15 @@ def make_arbitrary_grad(
         Orientation of gradient event of arbitrary shape. Must be one of `x`, `y` or `z`.
     waveform : numpy.ndarray
         Arbitrary waveform.
-    system : Opts, default=Opts()
+    system : Opts
         System limits.
-    max_grad : float, default=0
+        Will default to `pypulseq.opts.default` if not provided.
+    max_grad : float
         Maximum gradient strength.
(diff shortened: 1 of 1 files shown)


pypulseq/make_arbitrary_grad.py
```diff
@@ -45,36 +54,34 @@ def make_arbitrary_grad(
         If slew rate is violated.
         If gradient amplitude is violated.
     """
-    if system == None:
+    if system is None:
```
btasdelen, line 57: Changing defaults from 0 to None is I think more Pythonic, but also a breaking change if anyone was using these as 0, not that I think many people are using it. Would it make sense to accept both 0 and None in the checks below for some time and remove the check for 0 later?

btasdelen (commented):
Other than the single comment I made, the rest looks good.

pypulseq/make_arbitrary_grad.py
schuenke: I also think nobody ever sets a value to 0 in order to fall back to the defaults and I don't like it at all, but I agree that it is a breaking change. For the moment, I changed it to `if None or 0` and at some point we can maybe change it in all functions.

I also refactored the slew_rate calculation using `np.diff()`. Cleaner and more pythonic imo. Results are the same.

btasdelen (approved):

Issue #163, opened 2024-01-12
Feature: Sequence `install`


FrankZijlstra:
One main feature from pulseq that I personally use a lot at the scanner is the ability to install sequence files directly onto the host computer. In MATLAB pulseq this can be done with `seq.install('siemens')`, for example. I implemented a script to do this with pypulseq a while ago, but wasn't really happy with how Siemens-focused this code is. I made a llittle draft of how I thought we could be doing this in pypulseq in a bit more modular way: https://gist.github.com/FrankZijlstra/4049470aaaa7698d952c019e036796e7

The base `ScannerDefinition` class is the template for all `seq.install()` actions: We need to be able to check if the scanner is available (connected to the right network, etc.), and if so, be able to install a sequence file on it. Pypulseq would keep track of available scanner definitions (e.g. `SiemensDefinition` implements the Siemens-specific code that is available in MATLAB pulseq), but users would also be able to use `pp.register_scanner` to provide their own implementation (though ideally pypulseq would provide implementations for the cases where the `seq.install` workflow would be possible).

A scanner has a name, and optionally groups it belongs to, which can be used as the target for `seq.install(target)`. E.g. all Siemens definitions will be in a 'siemens' group, so if I'm at a Siemens scanner, but don't know which one, `seq.install('siemens')` would figure it out (more or less the behaviour of MATLAB pulseq). One feature that I do think is missing here is the option to bypass detecting the scanner, when I know for certain that everything works and I don't need to ping the scanner every time, maybe `seq.install('siemens_nx', fast=True)`?.

Does this make sense? I'm also curious if someone else has experience with pulseq on vendors other than Siemens, whether a similar `seq.install` workflow is even possible and could be implemented within this framework?

PR #190, opened 2024-08-15
`Sequence.install` functionality

FrankZijlstra:
This PR adds an implementation of the `Sequence.install` functionality as suggested in #163, with an interface that is very similar to Matlab Pulseq.
- The Matlab Pulseq implementation for Siemens scanners is included
- Custom scanner targets can be added using `pypulseq.Sequence.install.register_scanner`

The implementation is fairly straightforward, but I wonder if the `install` function should throw exceptions to give more detailed errors (e.g. stdout from the system command calls).


schuenke:
Didn't find the time yet, but I will hopefully review it soon

schuenke:
Hey. I finally went through the code. Sorry for the loooooooooong delay @FrankZijlstra 

I never used the Sequence.install() option, but hopefully be able to give it a try on Monday. 
Regarding the code, I think making `SequenceDefinition` a real [Abstract Base Class](https://docs.python.org/3/library/abc.html) would make sense. But I could do that if you want.

FrankZijlstra:
Yes, using the abstract base class makes sense. But since I never used it before, you are probably able to change it quicker than me. Having `can_install` default to `True` would be nice if possible, but it wouldn't be a big deal if that method is also abstract.

schuenke:
I tried to test the feature at our scanner, but in our institute the scanner is in a separate network so I cannot access it from my computer. 

The code in general looks good to me. Did you verify that it actually works as expected @FrankZijlstra? If you confirm it, I would be willing to approve it. Or does @btasdelen wanna give it a try?

FrankZijlstra:
Yes, I did test it (twice I believe), but only on one scanner (running Numaris X, not Numaris 4). The installation worked fine. At this point I don't remember if the caching of the scanner detection worked properly. I remember that once it kept pinging the scanner on every install call, despite being the same ipython runtime (i.e. where the cache should work). But I don't recall whether or not I fixed it afterwards... Either way that would be minor issue, that could be resolved easily later when someone is able to test it more thoroughly.

schuenke (approved):

PR #202, opened 2024-09-04
Fix direct install from GitHub repo

schuenke:
With #195 we switched from `setup.py` to `pyproject.toml`, where we defined:

```toml
[tool.setuptools.packages.find]
where = ["pypulseq"]
```

This works fine when you install pypulseq from a local folder using `pip install .`, but not if you want to install it directly from GitHub using `pip install git+https://github.com/imr-framework/pypulseq`.

With the direct install from GitHub, no `pypulseq` package will be created, but only the modules in our pypulseq folder (like SAR, seq_examples, Sequence, ...) will be installed as packages. 

For me, removing the 2 lines above solves the problem, which makes sense IMO, because this is meant to tell setuptools where to look for our package and not which folder IS our package. Setuptools detects the src-layout, where the pypulseq folder would be in a src folder (we should switch to this layout !!!) and the flat-layout (the one we are still using) automatically. This is why we don't have to specify where to look for our package at all.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,9 +35,6 @@ Homepage = "https://github.com/imr-framework/pypulseq"
 Issues = "https://github.com/imr-framework/pypulseq/issues"
 Documentation = "https://pypulseq.readthedocs.io/en/latest/"
 
-[tool.setuptools.packages.find]
-where = ["pypulseq"]
-
 [tool.setuptools.package-data]
 SAR = ["QGlobal.mat"]
 


FrankZijlstra:
Not an expert on this, but looks fine. Surprising that the behaviour is different depending on the install options. Have you checked whether it affects packaging the pypi package?

schuenke:
> Not an expert on this, but looks fine. Surprising that the behaviour is different depending on the install options. Have you checked whether it affects packaging the pypi package?

I didn't test it on PyPi directly, but tried the build command locally and installed it from the generated wheel in a clean env. That worked. Therefore, I am pretty sure it will work with PyPi as well.

btasdelen:
Yeah, sorry, apparently I misunderstood what it does, I thought it would add `pypulseq` as a package, too. 

Would you like to convert to `src` hierarchy in this PR, or should we go ahead and merge?

FrankZijlstra:
Moving everything to src might create big merge conflicts on the other PRs? I'd suggest to incorporate those first (and I have one more adding the sequence examples back to the sequence unit tests).

PR #210, opened 2024-12-03
Add MyPy type checker

schuenke:
This is still work in progress. Feel free to contribute 😁


schuenke:
I will close this for the moment. Enabling mypy on the current code base is ... let's say... difficult 😄 
If we do a major refactoring and switch to dataclasses at some point, we can give it another try...

Issue #213, opened 2024-12-04
make_trapezoid() creates gradient event with slew rate bigger than max_slew_rate without warning


h3lg3:
**Describe the bug**
Creating a gradient with make_trapezoid(), providing system, area, duration and rise_time creates a gradient with a slew rate bigger than system.max_slew without warning. 

**To Reproduce**
```
import pypulseq as pp
system = pp.Opts()

# Set max slew to 50 T/m/s
system.max_slew = 50 * system.gamma  # Hz/m/s

grad_spoil = pp.make_trapezoid(channel='z', system=system, area = 800, duration = 1e-3, rise_time = 200e-6)

grad_slew_rate = grad_spoil.amplitude/grad_spoil.rise_time/system.gamma
print(f"Slew rate of gradient: {round(grad_slew_rate)} T/m/s")

# Slew rate of gradient: 117 T/m/s
```

**Expected behavior**
If I set max_slew as a system spec I dont want it to be exceeded. 

**Desktop (please complete the following information):**
 - OS: [e.g. Win11]
- `pypulseq` version: [e.g. 1.4.2]

FrankZijlstra:
Resolved in #214.

PR #214, opened 2024-12-04
Assert that make_trapezoid() slew rate is smaller than max_slew

h3lg3:
Tries to address #213 

make_trapezoid() checks at the end if system.max_slew is exceeded. Similar to what is done with system.max_grad.

### DIFF
--- a/pypulseq/make_trapezoid.py
+++ b/pypulseq/make_trapezoid.py
@@ -199,7 +199,11 @@ def make_trapezoid(
 
     assert (
         abs(amplitude2) / rise_time <= max_slew
-    ), f'Refined slew rate ({abs(amplitude2)/rise_time:0.0f} Hz/m/s) is larger than max ({max_slew:0.0f} Hz/m/s).'
+    ), f'Refined slew rate ({abs(amplitude2)/rise_time:0.0f} Hz/m/s) for ramp up is larger than max ({max_slew:0.0f} Hz/m/s).'
+
+    assert (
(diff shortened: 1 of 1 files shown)


FrankZijlstra:
Yes, I agree that this should be checked. In addition, the same check should be done for the ramp down, i.e. using `fall_time` instead of `rise_time`.

And is it possible to create a gradient with a `rise_time` of 0.0 (and consequently, an amplitude of 0.0)? If so, the assertion needs to check that to avoid a division by 0.

@schuenke How do we deal with the ruff error in the pre-commit here? Should @h3lg3 run it locally to fix the formatting, or just change the formatting as suggested in the error log (https://github.com/imr-framework/pypulseq/actions/runs/12162263429/job/34029382533?pr=214)?

schuenke:
> @schuenke How do we deal with the ruff error in the pre-commit here? Should @h3lg3 run it locally to fix the formatting, or just change the formatting as suggested in the error log (https://github.com/imr-framework/pypulseq/actions/runs/12162263429/job/34029382533?pr=214)?

Ideally @sravan953 will activate pre-commit.ci for PyPulseq. If this is done, it would be fixed automatically. I asked him at GitHub and Slack, but didn't get a response yet. So for the moment it would be either @h3lg3 fixing it or one of us. Will probably take me 30 secs, so fine for me.

schuenke:
pre-commit.ci autofix

schuenke:
Okay, @sravan953 activated pre-commit ci and it seems to work fine.

PR #220, opened 2024-12-13
Add sequence / integration tests

schuenke:
This is still work in progress. More tests will be added.


FrankZijlstra:
I've attempted to add the example sequence tests (and some bugfixes to make the tests pass) to this PR. I also reworked the tests into a `TestSequence` class that avoids recreating the sequence multiple times (important for the more CPU-intensive sequence examples).

The tests run into an issue where it can't find the examples folder... Locally this works fine by adding the parent directory to `sys.path`, but on github this fails. Any thoughts @schuenke ? An alternative would be to move the sequence examples back into the main package...

Besides this:
- I think the `Approx` class should be taken out of `test_sequence.py`, it is a utility that could be of use in other tests (it extends pytest.approx to also work recursively on lists, dicts and SimpleNamespaces).
- I believe we should not ignore all warnings in the tests. For example, one of the sequence examples had timing errors, that now get checked by `write_seq` and give a warning. I'm not sure I understand the reason to give a `UserWarning` for the default duration of `make_block_pulse`. But it's easily avoided by specifying a duration.

schuenke:
> The tests run into an issue where it can't find the examples folder... Locally this works fine by adding the parent directory to `sys.path`, but on github this fails. Any thoughts @schuenke ? An alternative would be to move the sequence examples back into the main package...

The path problem should be fixed. Now it's "just" some AssertionErrors left. Maybe you can have a look again @FrankZijlstra? Reading your code comments I have the impression that you already spent some time on the write -> read issue 😄

FrankZijlstra (approved):

PR #222, opened 2024-12-17
Fix decimal places for uncompressed shapes

mavel101:
This PR fixes a problem, where uncompressed shapes get stored with too many significant digits. This could potentially lead to problems with floating point conversion and unnecessarily enlarges the file size.


FrankZijlstra:
I do not think this is necessary, and it introduces duplicate and/or premature rounding. When writing the sequence, shapes get written with 9 significant digits regardless of the internal precision.

The only problem I can see is that shapes that are identical within rounding errors will not get recognized as duplicates, but this is again caught before writing the sequence, and can be manually done by the user by calling `seq.remove_duplicates(in_place=True)`. But the same could happen in the `amplitude` field of a trapezoidal gradient, for example.

PR #227, opened 2024-12-30
Fix sigpy tests

btasdelen:
Sigpy RF pulse tests were throwing error. This is a hacky attempt to make them pass.


btasdelen:
@FrankZijlstra @schuenke A reminder about this PR. Maybe it is better to merge this with #220 ?

schuenke (approved):
I merged this with the latest changes and additional tests added in #237.
I am still not super happy with these tests in general, but I guess its fine for the moment.

PR #230, opened 2025-01-08
Fix None check in make_arbitrary_grad

schuenke:
Before this fix, it was not possible to set the `first` and/or `last` attribute to `0`, because `not 0` would always be true and thus the extrapolation was executed.


fzimmermann89 (approved):
👍 

(are there other cases where we convert 0/None to bool?)

PR #231, opened 2025-01-08
Merge dev into master

schuenke:
This PR is needed to update the new default branch (`master`) to the old one (`dev`)


btasdelen:
Should we do this after merging #212, #220, #227 ?

schuenke:
IMO we should have done it before switching the branches, but at least we should do it now as fast as possible.

If ppl install from GitHub, they currently get 2 years old code. This should be fixed asap.

btasdelen (approved):

schuenke:
@FrankZijlstra could you approve this as well? Need 2 approving reviews to merge into master now. 
There is no change compared to current dev branch. So no code to review.

FrankZijlstra (approved):

PR #232, opened 2025-01-10
Fix SAR calculation

### LINKED ISSUE #221: SAR calculation gives error due to missing RF signal
**Describe the bug**
Using the example sequence `write_epi.py` for the SAR calculation with the `pypulseq.SAR.SAR_calc.cal_SAR(seq)` function results in an error. 

`"SAR_calc.py", line 107, in _SAR_from_seq` 
`signal = rf.signal`
`AttributeError: 'NoneType' object has no attribute 'signal'`

**To Reproduce**
```
from write_epi import main as write_epi
import pypulseq as pp 

write_epi(plot = False, write_seq = True, seq_filename = 'epi_pypulseq.seq')

pp.SAR.SAR_calc.calc_SAR('epi_pypulseq.seq')
```

I would propose to check in addition to line 105 of `SAR_calc.py` if the signal of the rf event is not none.
Instead of 
```
if hasattr(block, "rf"):  # has rf
    rf = block.rf
    signal = rf.signal
```
use
``` 
if hasattr(block, "rf"):  # has rf 
    rf = block.rf
    if rf is not None:  # rf signal is not None
        signal = rf.signal
```

However, I have the feeling that the rf block is always present in the current version of pypulseq, in that case the first `if statement ` is obsolete, if that is not the case please correct me. 

Kind regards
Helge

schuenke:
Closes #221 

@h3lg3 and @joergfelder : would be great if you could test it.

### DIFF
--- a/src/pypulseq/SAR/SAR_calc.py
+++ b/src/pypulseq/SAR/SAR_calc.py
@@ -98,7 +98,7 @@ def _SAR_from_seq(seq: Sequence, Qtmf: np.ndarray, Qhmf: np.ndarray) -> Tuple[np
         block_dur = calc_duration(block)
         t[block_counter - 1] = t_prev + block_dur
         t_prev = t[block_counter - 1]
-        if hasattr(block, 'rf'):  # has rf
+        if hasattr(block, 'rf') and block.rf is not None:  # has rf event
             rf = block.rf
             signal = rf.signal
             # This rf could be parallel transmit as well


joergfelder:
Works fine! Thanks @schuenke

h3lg3:
Thank you @schuenke. Works for me too.

schuenke:
> Works fine! Thanks @schuenke

> Thank you @schuenke. Works for me too.

@btasdelen & @FrankZijlstra now I need you again to approve it 😄

FrankZijlstra (approved):

btasdelen (approved):

PR #235, opened 2025-01-14
Timing check of ADC delay

mavel101:
The timing check of the ADC currently uses the ADC raster time (typically 100ns) for all properties of the ADC (dwell time & delay).
While this raster time applies for the dwell time, it does not apply for the delay of the ADC, which must be on a 1us raster.
This issue can be fixed by setting "raster" to the RF raster time, as the dwell time of the ADC is anyway checked on the ADC raster time. I am not sure though, if this is the best solution style-wise.


mavel101:
Okay, I changed the respective line

schuenke:
We have to wait for #243 to be merged. I messed up the coverage path and might adjust the fail pipeline implementation a little bit. This is why the tests currently fail.

schuenke:
As explanation and reminder for the future:
grafik

schuenke (approved):
@mavel101 and @FrankZijlstra please let me know if my changes are fine for you so I can merge it.

mavel101:
Looks fine for me.

PR #237, opened 2025-01-17
Fix tests, update workflows, support Python 3.12 tests

schuenke:
- set rf.dead_time of refocussing pulse in write_epi_se (was causing an error after merge of #236). Sorry, my fault!
- cleaned push_pr workflow:
  - limit repository access to minimum
  - added Python 3.12 to matrix
  - included flag for sigpy tests in matrix (allows to run everything in parallel)
  - removed unnecessary use of conda env in tests
  - added dependency caching for speed up
  - added concurrency to save resources and avoid unnecessary duplication
  - added coverage report
  - ensure test job fails if Failures or Errors occur

All in all, this hopefully simplifies the testing process, removes unnecessary complexity, and is generally better structured.

### DIFF
--- a/tests/test_sigpy.py
+++ b/tests/test_sigpy.py
@@ -21,15 +21,12 @@ def test_sigpy_import():
 
 @pytest.mark.sigpy
 def test_slr():
-    import sigpy.mri.rf as rf
     from pypulseq.make_sigpy_pulse import sigpy_n_seq
 
-    print('Testing SLR design')
-
-    time_bw_product = 4
-    slice_thickness = 3e-3  # Slice thickness
+    slice_thickness = 3e-3
     flip_angle = np.pi / 2
-    # Set system limits
+    duration = 3e-3
+
     system = Opts(
         max_grad=32,
         grad_unit='mT/m',
@@ -49,10 +46,10 @@ def test_slr():
         band_sep=20,
         phs_0_pt='None',
     )
-    rfp, gz, _, pulse = sigpy_n_seq(
+    rfp, _, _ = sigpy_n_seq(  # type: ignore
         flip_angle=flip_angle,
         system=system,
-        duration=3e-3,
+        duration=duration,
         slice_thickness=slice_thickness,
         time_bw_product=4,
         return_gz=True,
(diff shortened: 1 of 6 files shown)


schuenke:
> The "slow" tests cannot run on their own, at least how it's currently set up, because it needs to load the sequence into the TestSequence object first with `test_sequence`, which is not marked as a "slow" test. A solution would be to change the class such that each test can create the sequence if it does not exist yet, or maybe we can mark `test_sequence` in a way that it always runs? (i.e. without -m, with -m "slow" and with -m "not slow"?).

Yeah, realized that as well. I introduced a "required_for_slow" marker and then use `pytest -m "slow or required_for_slow"` in the sequence tests. That seems to work. Will push this in a few minutes after some final tests.

schuenke:
Would be great if we can merge this soon, as we need it for #238 and #240 
@FrankZijlstra: I think you alread had a look at the code, no? Maybe you can review the latest changes and approve it if it's fine for you this way?!

tests/test_sigpy.py
```diff
@@ -64,29 +61,30 @@ def test_slr():
     seq = pp.Sequence()
```
FrankZijlstra, line 61: `system` should be passed to Sequence here.

tests/test_sigpy.py
```diff
@@ -123,15 +121,17 @@ def test_sms():
...
```
FrankZijlstra, line 121: `system` should be passed to `Sequence` here.

FrankZijlstra (approved):
Removing the `plot` tests for now is fine with me. Before adding the new Sequence tests, that was one of the only tests that existed, because @wtclarke found some bug with the plotting at the time, I think. But I agree that with automated tests in a headless environment, plotting could be difficult to get right...

FrankZijlstra:
> Why remove them? Can you not just patch the `show` method as I originally did? `@patch("matplotlib.pyplot.show")`. Should then be fine in the headless environment. You definitely want to test plotting code as it's something that users are likely to use and people are likely to fiddle with.

The plot tests have `show` patched (https://github.com/imr-framework/pypulseq/blob/4a25069569e863d110ac775bd5e65d1a25a99ea2/tests/test_sequence.py#L278). Apparently this was not enough to avoid @schuenke's error in his local environment.

The issue with plotting larger sequences (i.e. the sequence examples) is that it is very slow, so not ideal for automated tests on github.

schuenke:
> > Removing the `plot` tests for now is fine with me. Before adding the new Sequence tests, that was one of the only tests that existed, because @wtclarke found some bug with the plotting at the time, I think. It essentially just checks that calling `plot` raises no errors. But I agree that with automated tests in a headless environment, plotting could be difficult to get right...
> 
> Why remove them? Can you not just patch the `show` method as I originally did? `@patch("matplotlib.pyplot.show")`. Should then be fine in the headless environment. You definitely want to test plotting code as it's something that users are likely to use and people are likely to fiddle with.

I think we all agree that testing the plot function makes sense, but running an extremely slow seq.plot() for all example and test sequences without checking the output does not make sense IMO. Some decent tests for the plot function itself should be more useful.

wtclarke:
Surely just test on a small sequence then? I'd suggest adding a single block and try plotting that sequence. It's cool to see the tests over multiple sequence types, but don't use that as an excuse to not test a frequently used feature. As above, this is code that people will feel the need to tweak and users will use.

schuenke:
> Surely just test on a small sequence then? I'd suggest adding a single block and try plotting that sequence. It's cool to see the tests over multiple sequence types, but don't use that as an excuse to not test a frequently used feature. As above, this is code that people will feel the need to tweak and users will use.

I re-introduced and extended the seq.plot() tests for the basic test sequences (seq1, seq2, seq3, seq4). IMO this is still not a meaningful test, because we dont check the outcome at all, but now we know that the method does not raise any errors. Fine for you @wtclarke & @FrankZijlstra ?

wtclarke:
Thanks for doing this. Yeah, I understand your point, but I would say execution without error is still a meaningful test for something that is otherwise hard to test.

FrankZijlstra:
Yes, looks good to me. Minor nitpick is to put the `time_range` a bit higher, plotting the first millisecond is probably only ever going to plot the first block of the sequence.

PR #240, opened 2025-01-20
Fix grad.last calculation in read_seq function for arbitrary gradients

schuenke:
The extrapolation used to calculate grad.last was very buggy for arbitrary gradients leading to completely different results of seq.calculate_kspace() before/after writing the Sequence to a file and reading it in again.

Now the same extrapolation as in make_arbitrary_gradient is used. For my sequences, this fixed the problem

### DIFF
--- a/src/pypulseq/Sequence/read_seq.py
+++ b/src/pypulseq/Sequence/read_seq.py
@@ -262,7 +262,7 @@ def read(self, path: str, detect_rf_use: bool = False, remove_duplicates: bool =
         # We also need to keep track of the event IDs because some PyPulseq files written by external software may contain
         # repeated entries so searching by content will fail
         event_idx = self.block_events[block_counter]
-        # Update the objects by filling in the fields not contained in the PyPulseq file
+        # Update the objects by filling in the 'first' and 'last' attributes not yet contained in the Pulseq file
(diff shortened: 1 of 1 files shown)


schuenke:
Short reminder @btasdelen @FrankZijlstra 😁

btasdelen:
I tested this with my spiral trajectory. I basically designed the sequence, saved the trajectory using `seq.calculate_kspace()`. I also saved the trajectories after `read()` pre-fix and post-fix. After `read()` the trajectory is still not equal to design trajectory, but I am not sure if it is supposed to, considering the compression? Anyways, the difference between the trajectories is much smaller:

**Difference between designed and read trajectory before the fix:** 3486.3
**Difference between designed and read trajectory after the fix:** 887.7

Looks much better to me.

btasdelen (approved):

schuenke:
Thanks for the test. I did the same for my sequences before creating the PR. I think its never a perfect match with the current implementation due to the compression and rounding, but they should be very close already. 

I also designed a test for this, but for multi-echo and EPI sequences the results before/after writing to seq do not match yet. And we dont have a spiral traj in pypulseq directly. So I didnt include the tests here but might create another PR at some point where I also try to improve the calc kspace function.

PR #243, opened 2025-01-24
Fix coverage path and use pipefail

schuenke:
superseded by #246


fzimmermann89:
The coverage comment is no longer updated when the test fails

fzimmermann89:
Also, you can use
https://github.com/actions/runner-images/issues/4459#issuecomment-1374859960
and just set the shell to bash to get pipefail

schuenke:
superseded by #246

PR #244, opened 2025-02-03
Implemented rotate3D. Fix for rotate.

BenWilhelmUniklinikFreiburg:
Implemented rotate3D. Changed rotate to have the correct rotation direction when rotating around the y-axis. The direction for rotating around the x- and z-axis is not changed.

### DIFF
--- a/pypulseq/rotate3D.py
+++ b/pypulseq/rotate3D.py
@@ -1,0 +1,95 @@
+from types import SimpleNamespace
+from typing import List, Union
+
+import numpy as np
+from numpy import typing as npt
+
+from pypulseq.add_gradients import add_gradients
+from pypulseq.opts import Opts
+from pypulseq.scale_grad import scale_grad
+from pypulseq.utils.tracing import trace, trace_enabled
+
+
+def __get_grad_abs_mag(grad: SimpleNamespace) -> np.ndarray:
+    if grad.type == 'trap':
+        return abs(grad.amplitude)
+    return np.max(np.abs(grad.waveform))
+
+
+def rotate3D(*args: SimpleNamespace, rotation_matrix: np.ndarray[np.float64], system: Union[Opts, None] = None) -> List[SimpleNamespace]:
+    """
+    Rotates the corresponding gradient(s) by the provided rotation matrix. Non-gradient(s) are not affected.
+
+    See also `pypulseq.rotate.rotate()` and `pypulseq.Sequence.sequence.add_block()`.
+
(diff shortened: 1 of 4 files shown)


schuenke:
@BenWilhelmUniklinikFreiburg All source code needs to be in `src/pypulseq`. So you have to move the `rotate3D.py` file in order for the code to work and the tests to run.

BenWilhelmUniklinikFreiburg:
@schuenke sorry, not sure how that happened. The files should be in the correct place now.

FrankZijlstra:
Just a note: the `Approx` class in `test_sequence.py` should be able to handle checking the equality of arbitrary objects, as an extension of `pytest.approx`. So it can be used for checking the equality of gradient objects with tolerances. No need to write custom `compare_gradients` function for each test. Previously the `Approx` class was in a separate utils file in the tests folder, where I think it should be.

E.g. checking the equality of blocks works like this:
https://github.com/imr-framework/pypulseq/blob/1c623fe0ff7ad4908ccb6c634ae9c9ae33e0630d/tests/test_sequence.py#L319

mcencini:
Ok, scaling the area together with the other parameters in `scale_grad` fixes the inconsistency:

```
    scaled_grad = copy(grad)
        if scaled_grad.type == 'trap':
            scaled_grad.amplitude = scaled_grad.amplitude * scale
            scaled_grad.flat_area = scaled_grad.flat_area * scale
        else:
            scaled_grad.waveform = scaled_grad.waveform * scale
            scaled_grad.first = scaled_grad.first * scale
            scaled_grad.last = scaled_grad.last * scale
        scaled_grad.area = scaled_grad.area * scale
```

I think this is correct, isn't it? After all, it seems that `area` is a parameter for arbitrary `grad`  events - in other functions such as `add_gradients`, `area` is updated as well.

mcencini:
I opened a PR to @BenWilhelmUniklinikFreiburg fork with my changes - if merged, it should update this PR as well.

It also refactors tests so that it uses existing `Approx`, now moved to conftest.py to be re-used in different tests, as suggested by @FrankZijlstra.

As `Approx` was quite slow for `test_rotation3D_vs_rotation`, I refactored to speed it up. I later noticed that maybe `test_rotation3D_vs_rotation` generates unecessary long arbitrary waveforms (60000 points) - perhaps that is the reason for slow test runtime? If you like the older `Approx` version better, we can try reduce waveforms length and see if this is enough to achieve reasonable test runtime.

PR #246, opened 2025-02-05
Fix test workflow and enable workflow runs in PRs from forks

schuenke:
This PR currently blocks #235 and #240 , we should get it merged soon. 

The test workflow is now splitted into two seperate workflows:

1. Run all tests in the different environments and upload the coverage reports
2. Distinguish between PRs and pushes to master and create coverage comment in PRs 

Furthermore, always all tests including sigpy a run. However, I left the matrix solution for with / without sigpy in case we experience sigpy problems again in the future. In the future, we might also want to add a coverage badge to our readme. I added the required step for this and just commented it out for the moment.


.github/workflows/pytest.yml
```diff
@@ -0,0 +1,62 @@
...
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  test:
+    name: Code tests
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
fzimmermann89, line 15: why?

fzimmermann89 (approved):

PR #252, opened 2025-02-18
Fixes amplitude for RF refocusing gradient for write_tse,  issue #251

### LINKED ISSUE #251: Grad amplitude bug in example sequence write_tse.seq
The rf refocussing pulse gradient for slice selection uses the amplitude from the excitation pulse gradient. I believe the amplitude from the previously generated gz gradient for the refocussing pulse should be used instead. 

https://github.com/imr-framework/pypulseq/blob/0398b12bc1d86c1997ed5ad34242a22cc95003bd/examples/scripts/write_tse.py#L91 

```
    gs_ref = pp.make_trapezoid(
        channel='z',
        system=system,
        amplitude=gs_ex.amplitude, # here gz.amplitude should be used
        flat_time=t_refwd,
        rise_time=dG,
    )
```

h3lg3:
The rf refocussing pulse gradient for slice selection uses the amplitude from the excitation pulse gradient. I believe the amplitude from the previously generated gz gradient for the refocussing pulse should be used instead. Fixes #251


schuenke:
Hey Helge,

sorry for the suuuuper slow response on this. I had a quick look at the code and actually `gs_ex` is defined with `amplitude=gz.amplitude` so that `gs_ref.amplitude = gs_ex.amplitude = gz.amplitude` if I am not wrong. no ?!

Therefore, I would close the PR. Feel free to reopen it if I missed something.

PR #257, opened 2025-02-28
Fixes timing of GRE sequences

h3lg3:
The GRE sequence https://github.com/imr-framework/pypulseq/blob/0398b12bc1d86c1997ed5ad34242a22cc95003bd/examples/scripts/write_gre.py has a small TE/TR timing error if the rf.ringdown_time gets longer than the gz.fall_time. 
Taking the max() of both fixes the issue which becomes relevant for longer ringdown times or thick slices.

### DIFF
--- a/examples/scripts/write_gre.py
+++ b/examples/scripts/write_gre.py
@@ -60,14 +60,19 @@ def main(plot: bool = False, write_seq: bool = False, seq_filename: str = 'gre_p
     # Calculate timing
     delay_TE = (
         np.ceil(
-            (TE - pp.calc_duration(gx_pre) - gz.fall_time - gz.flat_time / 2 - pp.calc_duration(gx) / 2)
+            (
+                TE
+                - pp.calc_duration(gx_pre)
+                - max(gz.fall_time + gz.flat_time / 2, rf.ringdown_time + rf.shape_dur / 2)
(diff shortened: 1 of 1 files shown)


FrankZijlstra:
Going from the GRE sequence I wrote for https://github.com/pulseq/MR-Physics-with-Pulseq/, the correct delay calculation should be:
```delay_TE = (
    math.ceil(
        (
            TE
            - (pp.calc_duration(gz, rf) - pp.calc_rf_center(rf)[0] - rf.delay)
            - pp.calc_duration(gx_pre)
            - pp.calc_duration(gx) / 2
            - pp.eps
        )
        / seq.grad_raster_time
    )
    * seq.grad_raster_time
)```

This should fix your issue, and it does not assume that the center of the RF pulse is in the center of the waveform. `seq.test_report()` is accurate in reporting TE, so any errors with the nominal TE should show up there.

I had meant to update the example sequences based on the "clean" ones we developed for that course, but I never got around to it. This issue almost certainly exists in other examples.

h3lg3:
I adopted your proposed change. It seems that previously a rounding issue led to a delay_TE of 82 ms. Now it is 81 ms. Therefore the reference seq file for testing in '\tests\expected_output\' needs to be updated.

PR #264, opened 2025-03-10
Implement soft delay extension

btasdelen:
Work in progress. Effort to implement soft delay extension for file format 1.5.0 compatibility.

TODO: 
-------
- [ ] Add `soft_delay` handling to  `addBlock()`.
- [ ] Implement `apply_soft_delay()`.
- [ ] Add `soft_delay` handling to `write()`.
- [ ] Implement `get_default_soft_delay_values()`
- [ ] Add tests for soft delay.
- [ ] Add example sequence.

### DIFF
--- a/src/pypulseq/Sequence/sequence.py
+++ b/src/pypulseq/Sequence/sequence.py
@@ -32,6 +32,7 @@ from pypulseq.Sequence.ext_test_report import ext_test_report
 from pypulseq.Sequence.install import detect_scanner
 from pypulseq.Sequence.read_seq import read
 from pypulseq.Sequence.write_seq import write as write_seq
+from pypulseq.Sequence.write_seq import write_v141 as write_seq_v141
 from pypulseq.supported_labels_rf_use import get_supported_labels
 from pypulseq.utils.cumsum import cumsum
 from pypulseq.utils.tracing import format_trace, trace, trace_enabled
@@ -1730,7 +1731,12 @@ class Sequence:
         return all_waveforms
 
     def write(
-        self, name: str, create_signature: bool = True, remove_duplicates: bool = True, check_timing: bool = True
+        self,
+        name: str,
+        create_signature: bool = True,
+        remove_duplicates: bool = True,
+        check_timing: bool = True,
+        v141_compat: bool = False,
     ) -> Union[str, None]:
         """
         Write the sequence data to the given filename using the open file format for MR sequences.
@@ -1745,6 +1751,8 @@ class Sequence:
             Boolean flag to indicate if the file has to be signed.
         remove_duplicates : bool, default=True
             Remove duplicate events from the sequence before writing
+        v141_compat: bool, default=False
+            Write the sequence in v1.4.1 compatible file format.
 
         Returns
         -------
@@ -1783,7 +1791,10 @@ class Sequence:
                 warn(warn_msg, stacklevel=2)
 
         # Write the sequence
-        signature = write_seq(self, name, create_signature, remove_duplicates)
+        if v141_compat:
+            signature = write_seq_v141(self, name, create_signature, remove_duplicates)
+        else:
+            signature = write_seq(self, name, create_signature, remove_duplicates)
 
         # Return the sequence md5 signature if requested
         if signature is not None:
--- a/src/pypulseq/Sequence/sequence.py
+++ b/src/pypulseq/Sequence/sequence.py
@@ -1378,6 +1378,45 @@ class Sequence:
         self.extension_string_idx.append(extension_str)
         assert len(self.extension_numeric_idx) == len(self.extension_string_idx)
 
+    def apply_soft_delay(self, **kwargs):
+        """
+        Applies soft delays to the sequence by modifying the block durations of the respective blocks.
+        Input parameters are pairs of soft delays and values, whereas the soft delay is identified by its string hint and the value is the duration in seconds.
+        Not all soft delays defined in the sequence need to be specified.
+        Examples:
+           seq.apply_soft_delay(TE=40e-3) # set TE to 40ms
+           seq.applySoftDelay(TE=50e-3, TR=2) # set TE to 50ms and TR to 2 s
+
+        Parameters:
+        -----------
+        **kwargs : dict
+            Keyword arguments of the form (string hint, value) pairs
+            The string hint is the name of the soft delay to be modified.
+            The value is the new duration of the soft delay in seconds.
+        Raises:
+        -------
+        ValueError
+            If the given soft delay is not defined in the sequence.
+        ValueError
+            If the calculated delay is negative.
+        ValueError
+            If the string hint and numeric ID is inconsistent.
+        See Also
+        --------
+        pypulseq.make_soft_delay()
+        """
+        for _keys, _values in kwargs.items():
+            pass
+        # TODO: implement
+        return
+
+    def get_default_soft_delay_values(self) -> Tuple[bool, List[SimpleNamespace]]:
+        """
+        Go through all the blocks checking the consistency of the soft delays
+        TODO/FIXME
+        """
+        return
+
     def test_report(self) -> str:
         """
         Analyze the sequence and return a text report.
--- a/src/pypulseq/make_soft_delay.py
+++ b/src/pypulseq/make_soft_delay.py
@@ -1,0 +1,51 @@
+from types import SimpleNamespace
+
+
+def make_soft_delay(numID: int, hint: str, offset: float = 0.0, factor: float = 1.0) -> SimpleNamespace:
+    """
+    Creates a soft delay extension event.
+
+    Create a soft delay event, that can be used in combination with an empty (pure delay) block
+    e.g. to adjust TE, TR or other delays. The soft delay extension acts by rewriting the block duration
+    based on the user input (to the interpreter) according to the equation dur=input/factor+offset.
+    Required parameters are 'numeric ID' and 'string hint'. Optional parameter 'factor' can be either
+    positive and negative. Optional parameter 'offset' given in seconds can also be either positive and
+    negative. The 'hint' parameter is expected to be for identical 'numID'.
+
+    See Also
+    --------
+    - `pypulseq.Sequence.sequence.Sequence.add_block()`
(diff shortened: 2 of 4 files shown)


FrankZijlstra:
Regarding the failing sequence tests, you can set the environment variable SAVE_EXPECTED=1 and run pytest to locally rewrite the .seq files. This will automatically pass the write tests, but then allows you to do a diff with the previous .seq files (e.g. git diff, or make a copy of the .seq files first). Then if you manually verify that the change of IDs is expected, commit the new .seq files. I know manually checking this is a bit of a pain, but I think aiming for a exact match of the .seq files is a good standard, better safe than sorry.

src/pypulseq/Sequence/block.py
```diff
@@ -150,9 +155,22 @@ def set_block(self, block_index: int, *args: SimpleNamespace) -> None:
...
+                if hasattr(event, 'id'):
+                    event_id = event.id
+                else:
+                    event_id = register_soft_delay_event(self, event)
+                ext = {'type': self.get_extension_type_ID('DELAYS'), 'ref': event_id}
+                extensions.append(ext)
+            else:
+                raise ValueError(f'Unknown event type {event.type} passed to set_block().')
         else:
             # Floating point number given as delay
```
fzimmermann89, line 168: delay given as floating poiint number

src/pypulseq/Sequence/block.py
```diff
@@ -40,11 +40,16 @@ def set_block(self, block_index: int, *args: SimpleNamespace) -> None:
...
         If a gradient that doesn't end at zero is not aligned to the block boundary.
+        If multiple soft_delay extensions are used in a block.
+        If a soft delay extension is used in a block of zero duration.
+        If a soft delay extension is used in a block containing conventional events.
     """
     events = block_to_events(*args)
     new_block = np.zeros(7, dtype=np.int32)
     duration = 0
 
+    required_duration = None
```
fzimmermann89, line 51:
```suggestion
    required_duration:float|None = None
```

src/pypulseq/make_soft_delay.py
```diff
@@ -0,0 +1,51 @@
...
+def make_soft_delay(numID: int, hint: str, offset: float = 0.0, factor: float = 1.0) -> SimpleNamespace:
+    """
+    Creates a soft delay extension event.
+
+    Create a soft delay event, that can be used in combination with an empty (pure delay) block
+    e.g. to adjust TE, TR or other delays. The soft delay extension acts by rewriting the block duration
+    based on the user input (to the interpreter) according to the equation dur=input/factor+offset.
+    Required parameters are 'numeric ID' and 'string hint'. Optional parameter 'factor' can be either
+    positive and negative. Optional parameter 'offset' given in seconds can also be either positive and
+    negative. The 'hint' parameter is expected to be for identical 'numID'.
```
fzimmermann89, line 13: The 'hint' parameter is expected to be _identical_ (?) for identical 'numID'.

fzimmermann89 (commented):
I had a quick look -- and noticed i need more time too understand the logic and play around with the code....

fzimmermann89:
regarding check timings: for it to make sense, it would need to get the information about the values that should be used during the check, correct? maybe a dict of values?
 and check_timings should warn that the timings are only checked for this particular setting of delay values, other settings in the scanner UI can break.

src/pypulseq/Sequence/sequence.py
```diff
@@ -1433,6 +1487,204 @@ def set_extension_string_ID(self, extension_str: str, extension_id: int) -> None
...
+        """
+
+        # TODO: check if this works
+        # Go through all the blocks and update durations, at the same time checking the consistency of the soft delays
+        sd_str2numID = {}
+        sd_numID2hint = {}
+        sd_warns = {}
+        for block_counters in self.block_events:
+            block = self.get_block(block_counters)
+            if hasattr(block, 'soft_delay') and block.soft_delay is not None:
```
FrankZijlstra, line 1525: Add `soft_delay` to `attrs`  at top of `get_block` to always set it to `None`. Then remove the `hasattr` check here. Same in `get_default_soft_delay_values` below.

src/pypulseq/make_soft_delay.py
```diff
@@ -0,0 +1,51 @@
...
+
+    See Also
+    --------
+    - `pypulseq.Sequence.sequence.Sequence.add_block()`
+    - `pypulseq.Sequence.sequence.Sequence.apply_soft_delay()`
+
+    Parameters
+    ----------
+    numID : int
```
FrankZijlstra, line 22: I don't understand the reason the `numID` parameter exists and is not just assigned automatically, as is done for other events. If the `hint` needs to be identical for identical `numID`, then the delay can be uniquely addressed with just the `hint`.

btasdelen:
@FrankZijlstra @fzimmermann89 Many thanks for the comments. It looks like #288 takes this PR and has many changes on top of this. I am not sure how to proceed with this without duplicating effort and creating some conflicts.

It looks like there are some changes to soft delay part, but most changes on #288 are for other features. If @mcencini wants to adress these feedback on there I can close this PR.

src/pypulseq/make_soft_delay.py
btasdelen: I think the main purpose of the `numID` is letting the user choose which delay box they want to use for a specific delay. For example, if they want to put TE first then TR, they can do so by explicitly assigning IDs 0 and 1, to them, regardless of the order they are added into the sequence.

But, I agree with you. It creates confusion for little gain. Also, I tried setting `numID` to 2 and did not assign anything to 0 and 1, and interpreter shows 3 delays anyway since it has to create them sequentially. It gracefully handled empty events, but still.

I think we can deviate from the Matlab here.

src/pypulseq/Sequence/block.py
```diff
@@ -264,11 +296,18 @@ def set_block(self, block_index: int, *args: SimpleNamespace) -> None:
...
+        duration = required_duration
+
     self.block_events[block_index] = new_block
     self.block_durations[block_index] = float(duration)
 
 
-def get_block(self, block_index: int) -> SimpleNamespace:
+def get_block(self, block_index: int, add_ids: bool = False) -> SimpleNamespace:
     """
     Returns PyPulseq block at `block_index` position in `self.block_events`.
```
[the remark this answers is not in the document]
btasdelen: Good catch. I could not figure out the exact purpose yet, it looks like when true, it assigns an `id` member to all extensions. The only usage of this is I found in `Transform` functions: 
https://github.com/pulseq/pulseq/blob/43c281e5d99ca0bdd99b1334632ad2762527a2ef/matlab/%2Bmr/transform.m#L106
and
https://github.com/pulseq/pulseq/blob/43c281e5d99ca0bdd99b1334632ad2762527a2ef/matlab/%2Bmr/%40TransformFOV/TransformFOV.m#L359

For now, I will remove it since it is not relevant to soft delays. We can decide if we need it later on.

examples/scripts/write_gre_label_softdelay.py
```diff
@@ -0,0 +1,156 @@
...
+                channel='y',
+                area=phase_areas[i],
+                duration=pp.calc_duration(gx_pre),
+                system=system,
+            )
+            seq.add_block(gx_pre, gy_pre, gz_reph)
+            # seq.add_block(pp.make_delay(delay_TE))
+            seq.add_block(
+                10e-6,  # Add a small delay as we can't have soft_delay in an 0 duration block.
+                pp.make_soft_delay(numID=0, hint='TE', offset=-min_TE, factor=1.0),
```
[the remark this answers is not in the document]
btasdelen: I agree. For now, I removed the `required_duration`, now `default_duration` is an input to `make_soft_delay` with a default value of 10-e5.

> On a sidenote, I really think the float input as an option to add_block should not exist in the first place.

I looked into this, but I felt like removing that will cause bunch of other issues to solve: https://github.com/imr-framework/pypulseq/commit/99bbffd058d1284124a58a4c68b591205b147328

src/pypulseq/Sequence/block.py
btasdelen: `required_duration` is removed.

src/pypulseq/Sequence/block.py
```diff
@@ -13,7 +13,7 @@
 from pypulseq.utils.tracing import trace_enabled
 
 
-def set_block(self, block_index: int, *args: SimpleNamespace) -> None:
+def set_block(self, block_index: int, *args: Union[SimpleNamespace, float]) -> None:
```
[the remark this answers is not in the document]
btasdelen: Done.

FrankZijlstra:
I think it would be okay to merge it with a few minor changes:
- Add write_gre_label_softdelay to the example sequence tests in `test_sequence`. This checks basic things like writing and reading the sequence, which is good to test the soft delay extension blocks in the sequence file.
- Fix `get_default_soft_delay_values` so it does not contain the error report stuff. I do not fully understand the checks being done, but it seems to me it should be in `check_timing` (or maybe it can be checked immediately when the block gets added?). These can be left as placeholders wherever it makes most sense to put these checks. Note that `check_timing` automatically gets called before writing a sequence, so it is the most safe for any sequence checks that would breaks things on the scanner.

schuenke (approved):
I refactored the code quite a lot trying to take all previous comments into account. You can see a summary of all my changes using the GitHub comparison functionality:

https://github.com/imr-framework/pypulseq/compare/0374a60...40f17f5

The most important changes are:

- **Simplified API**: Made numID optional with automatic assignment based on hint order
- **Complete documentation**: Added comprehensive docstrings with examples and mathematical explanations  
- **Robust validation**: Improved error messages and parameter validation (including all whitespace types)
- **Comprehensive testing**: Added 14 test functions covering edge cases and boundary conditions
- **Better examples**: Created simple beginner example + improved complex example documentation
- **Code cleanup**: Removed broken functions, improved duration handling, cleaner float API separation

Would be great if @btasdelen @FrankZijlstra @mcencini and maybe @fzimmermann89 could have a look at it again. 

I am not sure about the soft_delay_simple_example, because it doesn't really fit into the available examples, but still thought it might be helpful. The documentation would probably be a good place for this, but IMO the docu is not the best atm and thus not really used by anyone...

PR #269, opened 2025-04-02
Add convert_to_arbitrary option to make_extended_trapezoid_area function

schuenke:
`make_extended_trapezoid_area` calls the `make_extended_trapezoid` function in the end to create the desired gradient. `make_extended_trapezoid` already supports the option to convert the resulting extended trapz gradient into an arbitrary gradient.

This PR simply introduces the option to set the `convert_to_arbitrary` flag in `make_extended_trapezoid_area` as well and parsing it to the `make_extended_trapezoid` function.

I kept the times and waveform arrays in the return values, although they are already available from the returned grad object. This became even clearer now due to the fact that I directly changed the return statement to `return grad, grad.tt, grad.waveform`
We might want to remove `grad.tt` and `grad.waveform` in the future, but I kept it in for backwards compatibility. 

I also added some tests checking that the area and duration with and without the `convert_to_arbitrary` flag match.

### DIFF
--- a/src/pypulseq/make_extended_trapezoid_area.py
+++ b/src/pypulseq/make_extended_trapezoid_area.py
@@ -217,11 +217,9 @@ def make_extended_trapezoid_area(
         times = cumsum(0, time_ramp_up, time_ramp_down)
         amplitudes = np.array([grad_start, grad_amp, grad_end])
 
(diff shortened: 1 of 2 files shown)


btasdelen (approved):
Looks straightforward. Merging.

PR #270, opened 2025-04-02
Consider floating point accuracy for system specs checks

h3lg3:
Due to rounding issue checks for violating system specs in `make_trapezoid()` may fail, even though max values are not exceeded. For example the test below fails. Adding machine precision to the comparison solves the issue and fixes #267 

```
abs(amplitude2) > max_grad
125000.00000000001 > 125000
# ValueError: Refined amplitude (125000 Hz/m) is larger than max (125000 Hz/m).
```

The error can be raised using this example script

```
from pypulseq import Opts
from pypulseq import make_trapezoid

system = Opts(
    rf_dead_time=20e-6, 
    rf_ringdown_time=30e-6,
    grad_raster_time=1e-6,
    rf_raster_time=1e-6,
    block_duration_raster=1e-6,
    adc_raster_time=1e-6,
    max_grad=250e3 / 42.576e6 * 1e3,
    grad_unit='mT/m',
    max_slew=50,
    slew_unit='T/m/s',
    B0=50e-3,
    gamma=42.576e6,
)

grad = make_trapezoid(
    channel='x',
    area=140.625,
    system=system,
    max_grad=125000.0,
)

print(grad.amplitude)  # 125000.00000000001
```

PR #272, opened 2025-04-03
Update to make_extended_trapezoid Slew Rate Violation

jbbacon:
The slew is calculated from the waveform and tt. After the differentiation these each have their own floating point errors which are the largest source of inaccuracy in calculating the slew, rather than its own intrinsic floating point error. The update therefore takes into account the size of the slew so it uses the absolute error based upon the calculation in line 131.


h3lg3:
Good point, this PR is linked to PR #270 and should be merged afterwards, or should be incorporated in PR #270.

btasdelen:
I am not sure if this is true. Diff operation only subtracts consecutive two elements, i.e. `out[i] = a[i+1] - a[i]`. So the error has nothing to do with the `max_slew`, only the two consecutive floats that are being subtracted has the effect. Scaling the `eps` by `max_slew` a) results in a huge `eps`, b) may cause much larger than intended slew rates to be accepted. Please correct me if I am missing something.

jbbacon:
Ah ok I think you are right that the formula I shared is not appropriate in this situation, but I still dont think it is as simple as just adding eps, I think there still needs to be some scaling. 

The reason I came across this was whilst trying to create a gradient with a slew of 200 T/m/s with a gradient raster time of 10e-6 s. After multiplying by the system gamma this gives an upper limit of 8,515,200,000 + 1e-9  to the slew, with the 1e-9 being the eps. 
My slew was being rejected for having a maximum of 8,515,200,000.000347. Upon investigation this was due to the division in line 131 being 85152.0/9.999999999999593e-06, instead of the requested 10e-6. So whilst both the values in line 131 are both reasonably within the eps limit, it resulted in a maximum slew which dramatically failed. This is why I made the change to account for the size of the slew being used as whilst the error is ~1e-4, the percentage error is ~1e-12.

btasdelen:
I see what you mean now. This is caused by very large floating point number being divided by very small floating time points. This is very similar to the issue from #114. As such, a proper fix would be converting the time axis to fixed point representation, but that would require an overhaul. The current solution might be okay for now.

Could you remove `.DS_Store` file before we can merge?

jbbacon:
Ah that makes sense, I believe the .DS_Store file should be removed

FrankZijlstra:
Would this inaccuracy also apply to other slew rate checks? E.g. in `make_arbitrary_grad` or `make_trapezoid` ?

btasdelen:
Yes, same check happens in those as well. @jbbacon would you mind making the same changes to other files as well? Locations I can find:

https://github.com/imr-framework/pypulseq/blob/9c57d9f362e98e8b5320286946975cca105b8135/src/pypulseq/make_trapezoid.py#L228-L239


https://github.com/imr-framework/pypulseq/blob/9c57d9f362e98e8b5320286946975cca105b8135/src/pypulseq/make_arbitrary_grad.py#L80

jbbacon:
No worries, those should all be in the PR now

btasdelen (approved):

PR #273, opened 2025-04-07
Fixes seq.plot() with labels issue #262

### LINKED ISSUE #262: plot with label
**Describe the bug**

when plot with label, there is a bug that "p is interactive object"
code write_gre_label.py
error on 
```python
seq.plot(label=['REP','LIN'],time_range=np.array([0, 3]) * TR, time_disp='ms')
```
I replace the sequence line 1054-1063
```python
  # Plot each label individually to retrieve each corresponding Line2D object
  # p = itertools.chain.from_iterable(
  #     [
  #         sp11.plot(__t, _lbl_vals, ".")
  #         for __t, _lbl_vals in zip(_t, lbl_vals)
  #     ]
  # )
  # if len(label_legend_to_plot) != 0:
  #     sp11.legend(p, label_legend_to_plot, loc="upper left")
  #     label_legend_to_plot = []
```
 by following
```python
  itertools.chain.from_iterable(
      [
          sp11.text(__t, 0.01+0.015*_count, _label_str+str(_lbl_vals))
          for __t, _lbl_vals,_label_str,_count in zip(_t, lbl_vals,label_legend_to_plot,range(len(lbl_vals)))
      ]
```
Also, replace line 999
```python
if valid_labels[i] in label.upper():
```
by 
```python
if valid_labels[i] in label_upper:
``` 
and add before it
```python
label_upper = [x.upper() for x in label]
```
Now it is working

h3lg3:
This PR fixes #262 
Labels are plotted correctly for seq.plot(label='lin'). 
Can be tested with "write_epi_label.py" and "write_gre_label.py"

### DIFF
--- a/src/pypulseq/Sequence/sequence.py
+++ b/src/pypulseq/Sequence/sequence.py
@@ -992,7 +992,7 @@ class Sequence:
                             [sp11.plot(__t, _lbl_vals, '.') for __t, _lbl_vals in zip(_t, lbl_vals)]
                         )
                         if len(label_legend_to_plot) != 0:
-                            sp11.legend(p, label_legend_to_plot, loc='upper left')
+                            sp11.legend(list(p), label_legend_to_plot, loc='upper left')
                             label_legend_to_plot = []
 
                 if getattr(block, 'rf', None) is not None:  # RF
--- a/src/pypulseq/Sequence/sequence.py
+++ b/src/pypulseq/Sequence/sequence.py
@@ -988,14 +988,12 @@ class Sequence:
                         t = t0 + adc.delay + (adc.num_samples - 1) / 2 * adc.dwell
                         _t = [t_factor * t] * len(lbl_vals)
                         # Plot each label individually to retrieve each corresponding Line2D object
-                        itertools.chain.from_iterable(
-                            [
-                                sp11.text(__t, 0.01 + 0.015 * _count, _label_str + str(_lbl_vals))
-                                for __t, _lbl_vals, _label_str, _count in zip(
-                                    _t, lbl_vals, label_legend_to_plot, range(len(lbl_vals))
-                                )
-                            ]
+                        p = itertools.chain.from_iterable(
+                            [sp11.plot(__t, _lbl_vals, '.') for __t, _lbl_vals in zip(_t, lbl_vals)]
                         )
+                        if len(label_legend_to_plot) != 0:
+                            sp11.legend(p, label_legend_to_plot, loc='upper left')
+                            label_legend_to_plot = []
 
                 if getattr(block, 'rf', None) is not None:  # RF
                     rf = block.rf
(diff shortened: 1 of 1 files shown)


btasdelen:
This is the plot we get with this patch:

image

as opposed to what we get if we just comment out offending `if len(label_legend_to_plot) != 0:` 

image

Is this really what we want? First result is not readable at all. If we wanted the label counters to be printed, I think:

1. We should still plot the dots, they are helpful in quickly glancing the evolution of the counters.
2. We should not type the label name (like `SLC`, `LIN`, etc.) on every point, we should only type the number. I am also not sure if this is desired, if y-axis is already telling the counter, isn't this just clutter?
3. We should fix the `if` clause and show the legend. Following is the only change we need for that:

```python
if len(label_legend_to_plot) != 0:
    sp11.legend(list(p), label_legend_to_plot, loc='upper left')
    label_legend_to_plot = []
```

The only change is `list(p)` for the first input to the `legend` function.

Let me know what you think.

h3lg3:
I agree that the first plot is not useful and cluttered. The second plot is useful. Maybe we can add a legend indicating which marker (green/red) indicates which label (LIN, ADC, ..)

src/pypulseq/Sequence/sequence.py
```diff
@@ -988,12 +988,14 @@ def plot(
...
-                            [sp11.plot(__t, _lbl_vals, '.') for __t, _lbl_vals in zip(_t, lbl_vals)]
+                        itertools.chain.from_iterable(
+                            [
+                                sp11.text(__t, 0.01 + 0.015 * _count, _label_str + str(_lbl_vals))
+                                for __t, _lbl_vals, _label_str, _count in zip(
+                                    _t, lbl_vals, label_legend_to_plot, range(len(lbl_vals))
+                                )
+                            ]
                         )
-                        if len(label_legend_to_plot) != 0:
```
btasdelen, line 994: I suggest we add these 3 lines back. Change `p` to `list(p)`.

src/pypulseq/Sequence/sequence.py
```diff
@@ -988,12 +988,14 @@ def plot(
                         t = t0 + adc.delay + (adc.num_samples - 1) / 2 * adc.dwell
                         _t = [t_factor * t] * len(lbl_vals)
                         # Plot each label individually to retrieve each corresponding Line2D object
-                        p = itertools.chain.from_iterable(
```
btasdelen, line 991: We restore this as well. Fixing the following if clause should solve the bug.

btasdelen (changes requested):
Let's restore this back and go from there.

src/pypulseq/Sequence/sequence.py
h3lg3: done.

btasdelen (approved):

PR #277, opened 2025-05-04
Update write_t2_se.ipynb

sairamgeethanath:
Had a typo in np.arange on line 258

### DIFF
--- a/examples/notebooks/write_t2_se.ipynb
+++ b/examples/notebooks/write_t2_se.ipynb
@@ -255,7 +255,7 @@
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


schuenke (approved):
I added some further improvements:

- remove contact info from about. We never do this in examples.
- adjust pip install command to new default branch
- rename some variables to be pep8 conform
- add delay and use arguments to rf pulses to show best practice and avoid warnings
- add system argument to adc

Please double check these changes before you merge it @sairamgeethanath

sairamgeethanath:
@schuenke - Thank you for scrubbing and cleaning up the notebook. We should undertake some effort to refactor examples in general as well, maybe after summer.

PR #278, opened 2025-05-05
Fix test report for X-nuclei sequences

schuenke:
The test report prints gradient and slew rate values in **mT/m** and **T/m/s**, respectively. In the `ext_test_report` function the `convert` function is used to convert values from **Hz/m** and **Hz/m/s**, but gamma was not provided. Therefore, the default value for 1H was used, which lead to wrong values in the test report for other nuclei then 1H.


sairamgeethanath:
Will take a look on next Monday.


*Sairam Geethanath, Ph.D.*
“There are plenty of compelling reasons to push MRI to be more powerful,
but that’s one facet of advancing the field. Another is delivering it to
the world.”

fzimmermann89 (approved):

PR #282, opened 2025-05-14
Fix:missing-t_ktraj-return-from-calculate_kspace

Nikbert:
This adds t_ktraj to the return values of calculate_kspace, to make it consistent with the Matlab Pulseq Version.
t_ktraj is very handy when comparing measured and nominal trajectories.

All the Best


FrankZijlstra:
I ran into this when I reimplemented `calculate_kspace` at the time. I agree that it is a useful return value, but any external script that uses `calculate_kspace` will break as a result.

One non-breaking option is to add a `return_t_ktraj` parameter, default to False, and give a deprecation warning when it is said to False (i.e. in some next release the old-style return values will be removed). When it is True, return the new return values as in the PR.

FrankZijlstra:
`calculateKspacePP` refers to the piecewise-polynomial implementation for k-space calculation. We already moved this to `calculate_kspace`, because it is the only implementation that should be used. `calculate_kspacePP` exists in pypulseq merely for backwards compatibility (it will give a deprecation warning when used), but should be removed at some point. So no, this function should not be touched.

Nikbert:
OK, then we can not mirror Pulseq calculateKspacePP. I think Franks suggestion sounds like a good solution. I think everyone that wants to match k-space data from a Skope fieldcam to the nominal trajectory would highly appreciate this change. I do not know how to alter this pull request accordingly. Frank can you help me?

btasdelen:
If we are going to introduce a breaking change with an option to roll back via a `return_XYZ` parameter, would it make sense to return a `dict` instead? This function is returning 5 values (6 with the new addition), not really clean.

FrankZijlstra:
I don't think it makes sense to have the rollback option not set as a default, because then it would still break old code unless the new parameter is set. With it set to default we can give a deprecation warning.

If we break things, I agree that returning a `dict` is nicer. Though a `SimpleNamespace` or result class (e.g. as sometimes used in scipy and numpy) is probably ever nicer. @schuenke probably can tell what is preferred if we do go that way.

To be clear, I'm not opposed to a breaking change, because even when we deprecate the current result values, at some point the breaking change needs to happen anyway. But this is something that would need to be documented. And a user should not get a "too many values to unpack" error without any indication why.

lukeje:
I also found I needed this parameter and so rolled up a quick solution in line with what Bilal suggested [here](https://github.com/lukeje/pypulseq/tree/calculate_kspace_dict) which seems to pass the standard tests. Is it worth me making a new pull request or merging it with this one somehow?

schuenke:
I will have a look at this again. Probably tuesday next week.

schuenke:
Since its been a while and I wasn't sure myself if everything was ported correctly, I let Claude do another review that might help whoever wants to review this PR:

Here you go:

---

## PR #282 Review: [calculate_kspace](cci:1://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/calculate_kspace.py:8:0-248:64) refactoring + `t_ktraj` return

I did a line-by-line comparison of the old inline method (master [sequence.py](cci:7://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/sequence.py:0:0-0:0) lines 283–458) against the new standalone [calculate_kspace.py](cci:7://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/calculate_kspace.py:0:0-0:0). **The code transfer is clean — no logic was lost.** All 195 relevant tests pass.

### Changes overview (5 files, +366 / −151)

| File | Change |
|---|---|
| [Sequence/calculate_kspace.py](cci:7://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/calculate_kspace.py:0:0-0:0) | **New file** — extracted code with helper functions, renamed variables, added `output_as_dict` |
| [Sequence/sequence.py](cci:7://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/sequence.py:0:0-0:0) | Thin wrapper delegating to `ext_calculate_kspace` |
| [Sequence/ext_test_report.py](cci:7://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/ext_test_report.py:0:0-0:0) | Updated to use `output_as_dict=True` |
| [tests/test_calculate_kspace.py](cci:7://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/tests/test_calculate_kspace.py:0:0-0:0) | **New file** — 3 tests for legacy, dict, and deprecated path |
| [tests/test_sequence.py](cci:7://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/tests/test_sequence.py:0:0-0:0) | 3 call sites updated to use `output_as_dict=True` |

### Line-by-line comparison: nothing lost

All logic was preserved with only variable renames for clarity:

| Old (master) | New (PR) |
|---|---|
| `gw_pp` | `grad_waveforms_pp` |
| `gm_pp` | `grad_moments_pp` |
| `ng` | `n_grad_channels` |
| `tc` | `time_candidates` |
| `ii` | `ramp_idx` / `idx_in_support` |
| `dk` | `k_offset` |
| `ii_next_excitation` | `next_excitation_idx` |
| `ii_next_refocusing` | `next_refocusing_idx` |
| `self.*` | `seq.*` (standalone function) |

The three helper functions ([_build_time_axis](cci:1://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/calculate_kspace.py:56:4-84:40), [_sample_gradient_moments](cci:1://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/calculate_kspace.py:86:4-107:21), [_apply_excitation_and_refocusing_resets](cci:1://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/calculate_kspace.py:109:4-152:56)) are clean extractions of the original inline code.

### Improvements found during transfer

1. **Bugfix in warning message** — The old code had `error={t_ktraj(i_period) - t_excitation(ii_next_excitation)}` using `()` (function call syntax) instead of `[]` (indexing). This would have **crashed with a TypeError** if the warning was ever triggered. The new code correctly uses `[]`.

2. **Removed dead commented-out code** — `# dk = -k_traj[:, i_period]` was cleaned up.

3. **Robustified empty-list handling** — Old `if tc != []:` replaced with `if len(time_candidates) > 0:` plus explicit `else: time_candidates = np.zeros(0)`. Functionally equivalent but cleaner.

### Things worth noting (not bugs, just observations)

1. **`slice_pos` is computed but never returned** — This exists in both old and new code. The MATLAB `calculateKspacePP` does return `slicepos`. Since `output_as_dict` was introduced to make the return value extensible, it would be natural to add `slice_pos` to the dict in a follow-up. Not a regression — this was pre-existing.

2. **[calculate_kspacePP](cci:1://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/sequence.py:333:4-347:102) correctly forwards `output_as_dict`** — The deprecated wrapper properly passes through the new parameter. ✓

3. **[ext_test_report.py](cci:7://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/ext_test_report.py:0:0-0:0) migration is correct** — Old unpacking [k_traj_adc, _, t_excitation, _, t_adc = self.calculate_kspace()](cci:1://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/src/pypulseq/Sequence/calculate_kspace.py:8:0-248:64) replaced with dict access. ✓

4. **[test_sequence.py](cci:7://file:///c:/Users/Patrick/MyDrive/Code/Python/pypulseq/tests/test_sequence.py:0:0-0:0) comparisons with `Approx`** — Switched from tuple to dict comparison. Works because `pytest.approx` supports dicts natively. ✓

### Verdict

**Safe to merge.** The refactoring faithfully preserves all original logic, the new `output_as_dict` feature is backwards-compatible (defaults to `False`), and the transfer actually fixed a latent bug in the warning message. The only optional follow-up would be including `slice_pos` in the dict output to eventually match MATLAB's `calculateKspacePP`.

schuenke (commented):
Since I did a lot of refactoring myself, I would suggest that somebody else reviews this PR as well.

mzaiss:
> Since I did a lot of refactoring myself, I would suggest that somebody else reviews this PR as well.

I would be happy to review it.

schuenke:
> > Since I did a lot of refactoring myself, I would suggest that somebody else reviews this PR as well.
> 
> I would be happy to review it.

Sounds good. Thanks.

PR #283, opened 2025-05-21
Fix get_gradients function for channel specific gradient delays

schuenke:
`Sequence.calculate_kspace()` as well as `Sequence.get_gradients()` (should) support the `trajectory_delay` argument to be of type `Union[float, List[float], np.ndarray]` to be able to define individual delays for the 3 different gradient channels. However, in `Sequence.get_gradients()` there was an additional multiplication with `ng` (the number of grad channels), leading to too long lists in the case of input type list or wrong values in the case of input type np.array. 

The lines below handle the gradient_offset, where the implementation (without the multiplication with ng) is correct already. 

In the [matlab code](https://github.com/pulseq/pulseq/blob/c4ffd8250bef0a6b6f75b9e1ad275409d11b7727/matlab/%2Bmr/%40Sequence/Sequence.m#L2012) the factor is also NOT present.

### DIFF
--- a/src/pypulseq/Sequence/sequence.py
+++ b/src/pypulseq/Sequence/sequence.py
@@ -270,10 +270,16 @@ class Sequence:
 
         Parameters
         ----------
-        trajectory_delay : float or list, default=0
+        trajectory_delay : float or list or numpy.ndarray, default=0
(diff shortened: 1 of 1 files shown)


fzimmermann89 (approved):
If you like, maybe add a sentence to the docstring what using a list vs a single numver means

PR #287, opened 2025-06-04
Integrate MRzero simulation into sequence class

mavel101:
This PR integrates the simulation of the complete sequence with MRzero (https://github.com/MRsources/MRzero-Core) similar to the Pulseq Matlab repo (https://github.com/pulseq/pulseq/blob/master/matlab/demoUnsorted/simMR0.m).
An example simulation (& FFT reconstruction) is added to the example GRE sequence.

### DIFF
--- a/src/pypulseq/Sequence/sim_mr0.py
+++ b/src/pypulseq/Sequence/sim_mr0.py
@@ -1,15 +1,15 @@
-from pathlib import Path
 import shutil
 import tempfile
 import urllib.request
+from pathlib import Path
 from typing import Optional
 
 import numpy as np
+
 from pypulseq import Sequence
 
 
 def sim_seq_mr0(seq: Sequence, work_path: Optional[str] = '', base_name: str = 'mr0sim', no_clean_up: bool = False):
-
     # Import inside function to avoid circular import
     import MRzeroCore as mr0
 


btasdelen:
I am not sure if it is a good idea to make `mr0` a dependency (even if optional) of PyPulseq by integrating it. `mr0` already has a good integration with PyPulseq, this feels like a circular dependency situation. In Pulseq, they just provided a simulation example code, they did not integrate it into the code base, which we can also do.

Even if we want to integrate this somehow, the current function provides very limited functionality, basically no option to change simulation parameters or select phantom. Again, I am totally fine for adding this as a standalone example/tutorial.

What are others opinions on this? @fzimmermann89 @schuenke @FrankZijlstra

mzaiss:
I also think with the full torch dependency of MR-zero it might get a bit large.

If you want to do it i recommend to use our util functions described below, which makes the calling simpler, even one liner possible:

signal, ktraj_adc = mr0.util.simulate(seq)  

https://github.com/MRsources/MRzero-Core/issues/125


it also allows us to maintain the standard values of simulation, if e.g.  accuracy logic changes.
You can also select different phantoms easier and we will further improve that.

I suggested the same  to matPulseq.

FrankZijlstra:
> signal, ktraj_adc = mr0.util.simulate(seq)

If that is available in MR-zero, I don't see a reason to include any special code in pypulseq.

p.s. good to see that that code got some use beyond the MR physics course :)

mzaiss:
An option would be to alter the toml file of pypulseq

[tool.poetry.extras]
mr0 = ["MRzeroCore"]


so in a CI test, or a interested user could install 

pip install pypulseq[mr0]

but this is still manual so not really a difference from separate installation.


(Yes, useful util things Frank! Your animations even made it to the Siemens Lunch Symposium at ISMRM)

fzimmermann89:
I also don't see a real advantage of having the mr0 simulation entrypoint in pypulseq.

 `pip install pypulseq[mr0]` +  `seq.sim_mr0` is not really different from  `pip install pypulseq mr0` +  `mr0.util.simulate(seq)`, is it?

But we should maybe include in both readme and documentation an example how to use mr0 to simulate the sequences, as it is quite useful to know about ;) 

Besides, i dont really like to have some function in pypulseq silently download a file from a different git repository and write it to disk..

Issue #258, opened 2025-03-01
Support for Pulseq v1.5.0


btasdelen:
Missing functionality to support new interpreter features:
- [ ] RF storage now includes RF center position, also supports frequency/phase offsets in PPM.
- [ ] ADC also supports frequency/phase offsets in PPM.
- [ ] Arbitrary gradients include first and last points.
- [ ] Soft delay extension.

Some functions/functionality that are not crucial for support but nice to have to catch up with Matlab Pulseq Toolbox.
- [ ] `calcAdcSeg()` function for long readouts.
- [ ] Demo sequences that use above features.
- [ ] `calcMomentsBtensor()`
- [ ] `paperPlot()` or simply improved plotting.
- [ ] Oversampling of arbitrary gradients.
- [ ] Scale gradients checks for system limits.

Tooine:
Hi, 

I'm using pypulseq to develop MR Fingerprinting sequences. While I was using those sequences only with the KomaMRI simulator until very recently (without problem), I can now access a real MR scanner to test them. Yet, the scanner interpreter is up to date and only accepts pulseq 1.5 sequences... 

So I was wondering when we might expect pypulseq to be compatible with pulseq 1.5? Depending on the availability, perhaps I should translate my code into Matlab. 

Also, I'm just curious to know if there exists a way to modify "by hand" a 1.4 pulseq sequence file to make it compatible for 1.5? 

Thanks!

schuenke:
Hi @Tooine 

In principle the Pulseq 1.5 interpreter should be compatible with Pulseq 1.4.x files. However, there is/was a bug in the code prohibiting the execution of older seq-files (including the gre_example sequence provided with the 1.5.0 interpreter code btw). Here is the corresponding Issue (https://github.com/pulseq/pulseq/issues/104) in the Pulseq repo. Maxim provided a fix in one of the comments: https://github.com/pulseq/pulseq/issues/104#issuecomment-2769385592

Tooine:
Thanks for the quick reply @schuenke! 

I have seen this issue on Pulseq but I though it was only fix in Matlab ... To what I understand now, the file `Arbitrary.cpp` in which we should make the modification is a file of the interpreter, not the (py)pulseq code, right? Can you confirm that? (sorry I don't have access to the interpreter myself so I'm just guessing as I did not find the file elsewhere)

fzimmermann89:
Yes, it is part of the interpreter.

fzimmermann89:
[...]

PR #291, opened 2025-06-19
[towards v1.5.x] Bump RF storage to v1.5.x

mcencini:
This PR partially addresses #258.

This only handles changes to RF storage format. The most substantial updates concerns `sequence.py` , `block.py` and `read_seq.py`/`write_seq.py` modules. Other changes are either very minor updates to include the new attributes in the pulse design routines (`make_*.py`) or are re-generation of the expected output for the test cases.

For the moment, I excluded a few tests failing due to some inconsistencies in gradient storage format when writing and reading again. I am planning to re-enabling them when bumping gradient storage format.

@btasdelen do you think this is small enough or should I try to further divide this? Thanks!

### DIFF
--- a/src/pypulseq/Sequence/sequence.py
+++ b/src/pypulseq/Sequence/sequence.py
@@ -1290,6 +1290,7 @@ class Sequence:
             'i': 'inversion',
             's': 'saturation',
             'p': 'preparation',
+            'o': 'other',
         }
         rf.use = use_cases.get(use, 'undefined')
 


mcencini:
Good idea! I guess we'll need multiple expected output files for the different storage formats

btasdelen:
I see what you mean. People who did not migrate their interpreters to 1.5 can stick with an older version of PyPulseq. I guess it boils down to the question, if we want people to use a newer version PyPulseq regardless, for them to have access to bug fixes and some of the new features.

We can also decide this down the line, probably.

src/pypulseq/Sequence/block.py
```diff
@@ -686,4 +702,7 @@ def register_rf_event(self, event: SimpleNamespace) -> Tuple[int, List[int]]:
     else:
         rf_id = self.rf_library.insert(key_id=0, new_data=data, data_type=use)
 
+    if hasattr(event, 'name'):
+        self.rf_id_to_name_map[rf_id] = event.name
```
btasdelen, line 706: I could not see this being used. Do we need this map? Event and later on `data` in the event library is already storing this information, no?

src/pypulseq/Sequence/sequence.py
```diff
@@ -1272,22 +1273,25 @@ def rf_from_lib_data(self, lib_data: list, use: str = str()) -> SimpleNamespace:
...
-            use_cases = {
-                'e': 'excitation',
-                'r': 'refocusing',
-                'i': 'inversion',
-                's': 'saturation',
-                'p': 'preparation',
-            }
-            rf.use = use_cases.get(use, 'undefined')
+        # TODO: fixme : use map built from pp.get_supported_rf_uses()
+        use_cases = {
```
btasdelen, line 1287: `other` is missing?

btasdelen (changes requested):
Just added two minor comments to be addressed, other than that looks good. Can be merged afterward.

src/pypulseq/Sequence/block.py
mcencini: Good question! I looked into MATLAB v1.5.0 release notes, it says `ability to add event names for display in the plots`.
Turns out they modified plotting, which now includes this block:

if isempty(rb.(field))
                out{3}=['\bf\color{blue}blk:\rm\color{black}' num2str(iB)];
            else
                try
                    switch field(1)
                        case 'a'
                            name = obj.hSeq.adcID2NameMap(rb.(field));
                        case 'r'
                            name = obj.hSeq.rfID2NameMap(rb.(field));
                        otherwise
                            name = obj.hSeq.gradID2NameMap(rb.(field));
                    end
                    out{3}=['\bf\color{blue}blk:\rm\color{black}' num2str(iB) ' \bf\color{blue}' field '\_id:\rm\color{black}' num2str(rb.(field)) ' ''\bf\color{darkGreen}' name '\rm\color{black}'''];
                catch 
                    out{3}=['\bf\color{blue}blk:\rm\color{black}' num2str(iB) ' \bf\color{blue}' field '\_id:\rm\color{black}' num2str(rb.(field))];
                end
            end
            
So I guess I'll have to use this map when it is time to update plotting tools. Should I remove the map from current PR?

btasdelen: Nope, it can stay if it will be used later on.

Issue #290, opened 2025-06-16
seq.plot() should return figure objects


J-Endres:
**Is your feature request related to a problem? Please describe.**
The option `seq.plot(plot_now=False)` exists so that the user can modify the figure before calling `plt.show()`.
Currently, this means retrieving the two used figures with `plt.figure(0)` and `plt.figure(1)`.
This can be error prone when at the time of calling other, user generated figures are open and the indices change.

**Describe the solution you'd like**
Change
https://github.com/imr-framework/pypulseq/blob/3dd64e038793a493e9912c7e522dc3e153a2a71b/src/pypulseq/Sequence/sequence.py#L1130-L1131
to
```python
if plot_now:
    plt.show()
else:
    return fig1, fig2
```
(and adjust the documentation / return value type hint)

**Describe alternatives you've considered**
Alternative: ignore this change. The problem can be circumvented easily on the user side, this feature would just make some applications sligthly easier.

schuenke:
In https://github.com/imr-framework/pypulseq/pull/308 we add a `clear` argument to the plot function and also adjust it to return the figures and subplot axes:

```
        Returns
        -------
        Tuple[plt.Figure, Tuple[plt.Axes, plt.Axes, plt.Axes], plt.Figure, Tuple[plt.Axes, plt.Axes, plt.Axes]]
            Returns (fig1, (sp11, sp12, sp13), fig2, (sp21, sp22, sp23)) for plot customization.
            Always returns figures and axes regardless of plot_now setting.
```

PR #296, opened 2025-07-11
[towards v1.5.x] Bump Plotting to v1.5.x 

mcencini:
This PR partially addresses https://github.com/imr-framework/pypulseq/issues/258

This updates plotting tools. Standard plot is now refactored in a custom class (similar to MATLAB), and also supports custom datatips (requiring an additional optional dependency), closing #290.

Also add paperPlot.


src/pypulseq/Sequence/sequence.py
```diff
@@ -917,221 +1001,13 @@ def plot(
             If false, plots are shown when plt.show() is called. Useful if plots are to be modified.
         plot_type : str, default='Gradient'
             Gradients display type, must be one of either 'Gradient' or 'Kspace'.
-        """
```
btasdelen, line 920: `-> None` it is returning `SeqPlot` now.

src/pypulseq/Sequence/sequence.py
```diff
@@ -13,26 +12,25 @@
 
     Self = TypeVar('Self', bound='Sequence')
 
-import matplotlib as mpl
 import numpy as np
-from matplotlib import pyplot as plt
 from scipy.interpolate import PPoly
 
 from pypulseq import __version__, eps
+from pypulseq.aux.paper_plot import paper_plot as ext_paper_plot
```
btasdelen, line 19: I think these should be put into `utils` rather than having a separate `aux` dir.

btasdelen (changes requested):
Just two minor comments, other than that everything looks good.

btasdelen (approved):

PR #308, opened 2025-09-08
Extend seq.plot() for reuse ability. (1.4)

### LINKED ISSUE #290: seq.plot() should return figure objects

mzaiss:
closes #290 

This PR makes seq.plot() more flexible for interactive use and debugging.
1. Added "clear=True" Parameter

The new clear flag defaulting to True ensures behavior as previously, which is that during repeated calls the plots in figure 1 and 2 get overwritten. Now, you can set clear=False to intentionally overlay sequences in the existing plots for direct comparison, which is useful for debugging subtle differences, but also great for educational purposes.

See here for two seqs with slightly different gradients (flat_area and flat_time vs. area and duration) and a corresponding shift in the rf pulse timing in this example.

SEQPLOT

2. Return Figure and Axes Objects

The function now returns its figure and axes objects (fig1, axes1, fig2, axes2), which allows for post-plot customization (e.g., adding titles, annotations, signal) or extracting axes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,14 +16,14 @@ repos:
       - id: mixed-line-ending
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.0
+    rev: v0.14.3
     hooks:
       - id: ruff # linter
         args: [--fix]
       - id: ruff-format # formatter
 
   - repo: https://github.com/crate-ci/typos
-    rev: v1.38.1
+    rev: v1.39.0
     hooks:
       - id: typos
         args: [] # empty, to remove write-changes from the default arguments.


mzaiss:
> Hi! I am not a maintainer, but just wanted to point out that [v1.5.0_dev](https://github.com/imr-framework/pypulseq/blob/v1.5.0_dev/src/pypulseq/utils/seq_plot.py) branch also has changes to plot.

i hope that the same reuse can be done there.
wil try to also make a PR

src/pypulseq/Sequence/sequence.py
```diff
@@ -990,15 +998,32 @@ def plot(
...
-        sp11 = fig1.add_subplot(311)
-        sp12 = fig1.add_subplot(312, sharex=sp11)
-        sp13 = fig1.add_subplot(313, sharex=sp11)
-        fig2_subplots = [
-            fig2.add_subplot(311, sharex=sp11),
-            fig2.add_subplot(312, sharex=sp11),
-            fig2.add_subplot(313, sharex=sp11),
-        ]
+        # Create the two figures (#1 for RF/ADC, #2 for gradients in x, y, z)
+        fig1 = plt.figure(1)
```
btasdelen: Is it a good practice to assign 1 and 2 as the figure IDs? Isn't it possible to have conflicts, as the user may have used these IDs before? Matplotlib starts assigning IDs from 1 and incrementing, so even if the user did not explicitly set it, figures may be overwritten.

btasdelen (changes requested):

Issue #268, opened 2025-04-01
PyQt5 requirement for Ubuntu/Linux systems


sairamgeethanath:
**Describe the bug**
seq.plot() does not display graphs

**To Reproduce**
On Ubuntu 22.04. LTS run write_gre.py with plot=True

**Expected behavior**
No plot shows

**Screenshots**
NA

**Desktop (please complete the following information):**
 - OS: Ubuntu 
 - OS Version: 22.04. LTS
- `pypulseq` version: 1.4.2

**Additional context**
Adding PyQt5 package solves the issue.

btasdelen:
This is more relevant to Matplotlib rather than PyPulseq. Matplotlib supports many "backends", PyQt5 is just one of them. Users should be able to pick whichever backend they want. Most Python installations come with Tkinter, so plots should be able to shown without installing any additional dependency using Tkinter. 

Documentation showing how to configure backends: https://matplotlib.org/stable/users/explain/figure/backends.html

If you configure matplotlib to use `TkAgg` using any of the recommended configuration options, you should be able to see the plots without installing PyQt.

sairamgeethanath:
Sure thing, best to have that as an option so that code does not break down when downloaded and used.

btasdelen:
I looked into this some time ago. I checked projects that are dealing with `matplotlib` as well. I could not see a nice way of handling this, without forcing a specific backend. 

I don't think calling `use()` inside `plot()` is a good solution. If the user or calls a plot with a different backend then `default_method` before calling our `plot()` function, it will also cause an error. Automatically setting a sane backend is already being done by `matplotlib`, and it usually fails when there is no available one, so I don't see how we can improve on that.

These are the two potential solutions I can think of:
1. We can provide an optional dependency on `PySide6` on `pyproject.toml`, then someone can install via `pip install pypulseq[Qt]`, for example, it will ensure an interactive backend is available.
2. We can check if the active backend is non-interactive and print a warning (I believe matplotlib already should print such a warning, but in case it is not clear):
```python
import matplotlib

current_backend = matplotlib.get_backend()
print(f"Current Matplotlib backend: {current_backend}")

non_interactive_backends = ['agg', 'pdf', 'ps', 'svg', 'cairo']

if current_backend.lower() in non_interactive_backends:
    print("A non-interactive backend is currently selected.")
else:
    print("An interactive backend is currently selected.")
```
#323 might be a good PR to add this check.

[...]

PR #328, opened 2025-11-11
Bump to v1.5.0

mcencini:
This PR fix conflicts with recent changes to master. These mainly regards plot (only other change concerns mod_grad_axis, which I updated to match the new internal gradient representation).

Specifically, seq.plot() now returns a "SeqPlot" object, which contains the figure handles. In addition to enable re-usability like #308 did, it improves interactivity (which requires the optional mpcursors package, added as optional dependency). Some examples below:

**Standard behavior**
The standard seq.plot() behaviour is preserved:

StandardPlot

Code also supports the overlay of multiple plots, as

```Python
fig = seq1.plot(time_range=[0.0, 0.012], plot_now=False)
seq2.plot(time_range=[0.0, 0.012], overlay=fig)
```

resulting in:

Overlaid

**New features**
In addition, following MATLAB toolbox, this class now supports stacked plots:

```Python
seq.plot(time_range=[0.0, 0.012], stacked=True)
```

StackedPlot

When `mplcursors` is installed, we can also datatips by clicking on the waveforms:

Datatips

we can also add a moving hairline following the cursor by setting `show_guides=True`:

```Python
seq.plot(time_range=[0.0, 0.012], show_guides=True)
```
StandardPlotMovingHairline

which also works when `stacked=True` (not shown).

The code is also compatible with the recent SoftDelay extension:

SoftDelay

For the moment, I disabled tests. I would like to discuss it before re-enabling (@schuenke): this version replace the optional fig1, fig2 arguments of seq.plot() with "overlay" (a SeqPlot object). In some of the test case, we create the empty figure handles and then fill it in plot(). Do we really want this feature? I figured that the "overlay" argument is more oriented towards overlaying existing plots, rather than filling empty handles. Current SeqPlot has fig1, fig2, ax1 and ax2 as attributes, so user can still access them and customize it if they need to. What do you think?  

EDIT: I also added an optional PySide6 dependency to address #268 @btasdelen

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ PyPulseq enables vendor-neutral pulse sequence design in Python [[1,2]][section-
 exported as a `.seq` file to be run on Siemens, [GE], [Bruker] and now also Philips hardware by leveraging their respective Pulseq interpreters. This tool is targeted at MRI pulse sequence designers, researchers, students and other interested
 users. It is a translation of the Pulseq framework originally written in Matlab [[3]][section-references].
 
-👉 Currently, PyPulseq is compatible with Pulseq >= 1.4.0. The new features introduced with Pulseq 1.5.0 are not yet supported, but will be added in the near future. 👈
+👉 Currently, PyPulseq is compatible with Pulseq >= 1.5.0. The new features introduced with Pulseq 1.5.1 are not yet supported, but will be added in the near future. 👈
 
 It is strongly recommended to first read the [Pulseq specification]  before proceeding. The specification
 document defines the concepts required for pulse sequence design using PyPulseq.


schuenke:
I don't have the perfect overview over the different branches on the different forks, but always updating the dev branch after master was updated would have been the easiest and cleanest way I guess. This would better separate your modifications from other modifications and thus simplify the review process. 

Independent of the branch structure, I realized that for example the changes from #308 are not taken into account in your code. It's not obvious, because the plot function was moved into a separate file, but for example the unification of the subplot naming and handling is completely reverted, same for the plot_type parameter, which appears again in your docstring although it's not a parameter anymore. Would also be great to have the init parameters documented in the `__init__ ` method instead of the class docstring itself. Maybe you can have a look at this again and try to merge the changes from #308 into your code. 

I am happy to also review everything afterwards. 

And don't get me wrong please: IMO it's great work that you have been doing (!!!), but we should make sure that we include all previous adjustments and try to steadily increase the code and documentation quality.

src/pypulseq/Sequence/block.py
```diff
@@ -254,7 +255,7 @@ def set_block(self, block_index: int, *args: Union[SimpleNamespace, float]) -> N
                     if prev_type == 't':
                         last = 0
                     elif prev_type == 'g':
-                        last = prev_lib['data'][5]
+                        last = prev_lib['data'][2]  # v150
```
schuenke: I would either be more precise and document what was changed. e.g. "`v1.5.0: changed from ['data'][5] to ['data'][2]`", or remove the comment. 

I prefer option 1, but highly recommend a uniform style for all these comments e.g. `v1.5.0: <text>,` which allows us to remove these more easily in the future.

src/pypulseq/Sequence/block.py
```diff
@@ -273,7 +274,7 @@ def set_block(self, block_index: int, *args: Union[SimpleNamespace, float]) -> N
                     if next_type == 't':
                         first = 0
                     elif next_type == 'g':
-                        first = next_lib['data'][4]
+                        first = next_lib['data'][1]  # v150
```
schuenke: see my comment about a uniform comment style above

src/pypulseq/utils/seq_plot.py
```diff
@@ -0,0 +1,686 @@
...
+    ax1 : tuple of matplotlib.axes.Axes
+        Tuple of axes for fig1: (sp11, sp12, sp13) if not stacked, or (sp11, sp12, sp13, sp21, sp22, sp23) if stacked.
+    ax2 : tuple of matplotlib.axes.Axes
+        Tuple of axes for fig2: (sp21, sp22, sp23) if not stacked, or same as ax1 if stacked.
+    vlines : dict (axis -> Line2D) if show_guides enabled
+    """
+
+    def __init__(
+        self,
+        seq,
```
schuenke: type hint missing

src/pypulseq/Sequence/block.py
```diff
@@ -496,25 +576,52 @@ def register_adc_event(self, event: EventLibrary) -> int:
...
+        # Clear block cache if overwritten
+        if self.use_block_cache and found:
+            self.block_cache.clear()
+
+    # Optional mapping
+    if hasattr(event, 'name'):
+        self.adc_id_to_name_map[adc_id] = event.name
 
-    return adc_id
+    return adc_id, shape_id
```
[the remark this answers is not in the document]
schuenke: The docstring is still off:

1. the return type should be `int, int` or `Tuple[int, int]` instead of `int[int, ...]`
2. it must be `ID of registered ADC` instead of `ID of registered RF`
3. `shape IDs` should be `shape ID`, no? Or do I misunderstand the code and it could actually be multiple shape IDs ?

src/pypulseq/calc_rf_bandwidth.py
```diff
@@ -1,17 +1,21 @@
...
+from pypulseq.opts import Opts
 
 
 def calc_rf_bandwidth(
     rf: SimpleNamespace,
     cutoff: float = 0.5,
     return_axis: bool = False,
     return_spectrum: bool = False,
+    dw: float = 10,
+    dt: float = 1e-6,
```
schuenke: better use the default system.rf_raster_time here instead of hardcoding a value?
probably best practice to make it float | None with default None and if None, we use the default?

src/pypulseq/make_adc.py
```diff
@@ -16,6 +18,9 @@ def make_adc(
     freq_offset: float = 0,
     phase_offset: float = 0,
     system: Union[Opts, None] = None,
+    freq_ppm: float = 0,
+    phase_ppm: float = 0,
+    phase_modulation: np.ndarray = None,
```
schuenke: Should be `Optional[np.ndarray]` or even better `Union[np.ndarray, None]` , which is the (ugly) default syntax we currently use. 

In the future, I would recommend to switch to `np.ndarray | None` syntax , which however requires Python > 3.10. So we have to discuss that first.

src/pypulseq/make_arbitrary_rf.py
```diff
@@ -63,8 +67,14 @@ def make_arbitrary_rf(
         System limits.
     time_bw_product : float, default=4
         Time-bandwidth product.
-    use : str, default=str()
+    use : str, default='undefined'
         Use of arbitrary radio-frequency pulse event. Must be one of 'excitation', 'refocusing' or 'inversion'.
```
schuenke: The default is not listed as valid option. The valid options listed in get_supported_rf_uses() are:
'excitation', 'refocusing', 'inversion', 'saturation', 'preparation', 'other', 'undefined'

src/pypulseq/make_arbitrary_grad.py
```diff
@@ -76,28 +80,59 @@ def make_arbitrary_grad(
     if channel not in ['x', 'y', 'z']:
         raise ValueError(f'Invalid channel. Must be one of x, y or z. Passed: {channel}')
 
-    slew_rate = np.diff(waveform) / system.grad_raster_time
+    if first is None:
+        # warnings.warn(
+        #     'it will be compulsory to provide the first point of the gradient shape in the future releases; finding the first by extrapolation for now...',
+        #     FutureWarning,
+        # )
```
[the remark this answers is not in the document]
schuenke: I leave this conversion unresolved for the moment so that we can further discuss it with @btasdelen @sairamgeethanath and @FrankZijlstra

src/pypulseq/utils/paper_plot.py
```diff
@@ -0,0 +1,121 @@
...
+    rf_plot : {'abs', 'real', 'imag'}, default='abs'
+        Determines how to plot RF waveforms (magnitude, real or imaginary part).
+
+    """
+    # Get waveform data
+    wave_data, _, _, t_adc, _ = seq.waveforms_and_times(append_RF=True, time_range=time_range)
+
+    # Max amplitudes for scaling
+    gwm = np.max(np.abs(np.concatenate(wave_data[:3], axis=1)), axis=1)
+    gwm[0] = max(gwm[0], t_adc[-1])
```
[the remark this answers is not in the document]
schuenke: I dont think we have to catch the empty sequence. 

Your change fixed the issue for me 👍

src/pypulseq/make_arbitrary_grad.py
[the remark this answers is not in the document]
schuenke: Fine for me 👍

src/pypulseq/make_arbitrary_rf.py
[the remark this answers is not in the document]
mcencini: Perfect, thanks!

mcencini:
> > P.S: I am not sure if it slipped due to the many changes of this PR or you are not arrived to that point yet, but for the moment the test_plot you wrote for #308 are commented out as the plot() signature slightly changed (it does not accept figure handles directly anymore). I just want to check in with you whether the current interface is ok before attempting to adapt it myself, as you are the original author of the tests. Thanks again!
> 
> @mcencini I reimplemented some of the tests in [d5a4d19](https://github.com/imr-framework/pypulseq/pull/328/commits/d5a4d19d79113bdcbbcf6f799895960bfc46cf2d). The others and not needed anymore imo so it's fine for me as it is.
> 
> I might want to refactor the whole sequence tests in the future using pytest fixtures and individual test functions instead of the TestSequence class, but this would blow up this PR even more so I am postponing it for the moment.

Perfect, thank you very much!

sairamgeethanath:
I am back on Tuesday and will get back to this in the evening for sure.
Thanks all!

*Sairam Geethanath, Ph.D.*
“There are plenty of compelling reasons to push MRI to be more powerful,
but that’s one facet of advancing the field. Another is delivering it to
the world.”

PR #337, opened 2026-01-12
Revert "Bump to v1.5.0"

sairamgeethanath:
Reverts imr-framework/pypulseq#328


schuenke (changes requested):
I am against merging this. See my comment above.

PR #340, opened 2026-01-12
Fix write_v141 function

ckolbPTB:
The pypulseq version in write_v141 needs to be hardcoded otherwise it is set to 1.5.0.

### DIFF
--- a/src/pypulseq/Sequence/write_seq.py
+++ b/src/pypulseq/Sequence/write_seq.py
@@ -327,6 +327,8 @@ def write_v141(self, file_name: Union[str, Path], create_signature, remove_dupli
         output_file.write('revision 1\n')
         output_file.write('\n')
 
+        ppm_to_hz = 1e-6 * self.system.gamma * self.system.B0
+
         if len(self.definitions) != 0:
             output_file.write('[DEFINITIONS]\n')
             keys = sorted(self.definitions.keys())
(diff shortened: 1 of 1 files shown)


mcencini (dismissed):
Good catch! Looks good to me

src/pypulseq/Sequence/write_seq.py
mcencini: This should (better to double check) be:
```suggestion
s = id_format_str.format(
k,
self.grad_library.data[k][0],
*self.grad_library.data[3:5],
round(self.grad_library.data[k][5] * 1e6),
)
```

schuenke:
I think I adjusted the write function to the new libray structures. For the sequences I checked, it was correct. 
@ckolbPTB can you try to use it in the [MRseq](https://github.com/PTB-MR/mrseq) workflows?

And should we maybe rename the function / flag to v142 instead of v141? 1.4.2 was the last version using the "old" file format, no?

src/pypulseq/Sequence/write_seq.py
```diff
@@ -377,8 +377,8 @@ def write_v141(self, file_name: Union[str, Path], create_signature, remove_dupli
             id_format_str = '{:.0f} {:12g} {:.0f} {:.0f} {:.0f} {:g} {:g} {:g}\n'  # Refer lines 20-21
             for k in self.rf_library.data:
                 lib_data1 = self.rf_library.data[k][0:4]
-                lib_data2 = self.rf_library.data[k][5:7]
-                delay = round(self.rf_library.data[k][4] / self.rf_raster_time) * self.rf_raster_time * 1e6
+                lib_data2 = self.rf_library.data[k][8:10]
```
mcencini: This is ignoring freq_ppm and phase_ppm. Maybe it is better to convert the ppm offsets to actual offsets using the stored` seq.system.B0`.

src/pypulseq/Sequence/write_seq.py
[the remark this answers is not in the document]
mcencini: No worries! :D I double checked and event_lib.data should be a numpy array, so += should indeed do elementwise sum rather than concatenation, and if I am not wrong the expression matches the [MATLAB](https://github.com/pulseq/pulseq/blob/28c55ea9050f93c28faa032185527459c29d69c8/matlab/%2Bmr/%40Sequence/write_v141.m#L72) counterpart

mcencini (approved):
To me it looks good to be merged as soon as @ckolbPTB is happy with that. Optionally, we could update the hardcoded version to 1.4.2 and the name to `write_142` as discussed - this would clearly reflect that this is compatible with latest 1.4.x revision, but also diverge from MATLAB

schuenke:
> To me it looks good to be merged as soon as @ckolbPTB is happy with that. Optionally, we could update the hardcoded version to 1.4.2 and the name to `write_142` as discussed - this would clearly reflect that this is compatible with latest 1.4.x revision, but also diverge from MATLAB

Actually, I want to correct myself and suggest to stay with 1.4.1 instead of 1.4.2 because older file versions are compatible with newer interpreter versions, but not the other way around. So we should pick the LOWEST possible version. If MATLAB uses 1.4.1, we should do the same.

PR #345, opened 2026-01-19
Fix backward compatibility 

ckolbPTB:
I run into several problems when I try to read in v141 files. 

I added a test and started fixing some problems but now I get the following error when trying one of my old seq files:
File ~/pypulseq/src/pypulseq/Sequence/block.py:481, in get_block(self, block_index)
    479 adc.phase_ppm = lib_data[4]
    480 adc.freq_offset = lib_data[5]
--> 481 adc.phase_offset = lib_data[6]
IndexError: tuple index out of range

which is different from the error in the test. 

Would it make sense to create a seq file of one simple sequence (write_gre?) for all the versions which pypulseq is supposed to support and add it to the repo for testing? This would at least allow for some basic testing of the read and write backward compatibility.

### DIFF
--- a/src/pypulseq/Sequence/read_seq.py
+++ b/src/pypulseq/Sequence/read_seq.py
@@ -56,7 +56,8 @@ def read(self, path: str, detect_rf_use: Union[bool, None] = None, remove_duplic
     # Raster times
     self.grad_raster_time = self.system.grad_raster_time
     self.rf_raster_time = self.system.rf_raster_time
-
+    self.adc_raster_time = self.system.adc_raster_time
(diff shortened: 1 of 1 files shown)


ckolbPTB:
Maybe having this part of github actions might be cleaner. Happy to give it a try if you want! Which versions should be supported?

src/pypulseq/Sequence/read_seq.py
mcencini: See above

```suggestion
    self.adc_raster_time = self.system.adc_raster_time
    self.block_duration_raster = self.system.block_duration_raster
```

PR #366, opened 2026-02-11
harmonize and parametrize example sequences

mzaiss:
The pypulseq example scripts are frequently used as de-facto tests. 
I realized certain dviation and have now harmonized them to follow a consistent pattern. The changes are designed to be fully transparent: existing behavior is preserved, and any current usage that relies on input parameters instead of variables continues to work without modification. At the same time, the refactoring enables testing with variable parameter sets by calling the functions directly, without violating backward compatibility.

FOV def always set
variable as default input

same structure for mprages as for others  (this could probably break something, as it wasn't functions here before)
fixed raster time issues in mprage 

In a second (optional) commit I suggested changing the FOV for multislice seqs for being consistent slice_thinckness*n_slices

All sequences were tested in pypulseq 1.4.2 (only)

### DIFF
--- a/examples/scripts/STYLE_GUIDE.md
+++ b/examples/scripts/STYLE_GUIDE.md
@@ -52,7 +52,7 @@ def main(
     write_seq: bool = False,
     seq_filename: str = 'gre_pypulseq.seq',
     *,
-    fov: float = 256e-3,
+    fov: float | tuple[float, float] = 256e-3,
     n_x: int = 64,
     n_y: int = 64,
     flip_angle_deg: float = 10,
@@ -80,8 +80,9 @@ the return value:
         Write the sequence to a .seq file. Default is False.
     seq_filename : str, optional
         Output filename for the .seq file. Default is 'gre_pypulseq.seq'.
-    fov : float, optional
-        Field of view in meters. Default is 256e-3.
+    fov : float or tuple of float, optional
+        Field of view in meters. If a single value, it is used for both x and y.
+        If a tuple, it is (fov_x, fov_y). Default is 256e-3.
     n_x : int, optional
         Number of readout samples. Default is 64.
     ...
@@ -145,7 +146,9 @@ if __name__ == '__main__':
 | `gx_spoil` | Readout spoiler gradient |
 | `gz_spoil` | Slice spoiler gradient |
 | `adc` | ADC readout event |
-| `delta_k` | k-space step size (`1 / fov`) |
+| `delta_kx` | k-space step size in readout direction (`1 / fov_x`) |
+| `delta_ky` | k-space step size in phase encoding direction (`1 / fov_y`) |
+| `delta_kz` | k-space step size in partition/slice direction (`1 / fov_z`, 3D only) |
 | `phase_areas` | Array of phase encoding areas |
 
 ### Derived / Timing Variables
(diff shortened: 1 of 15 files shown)


mcencini:
Yes, sure, I also think it is better to proceed incrementally - I just thought it would be a good spot to bring this up :)

mcencini:
Regarding failing test, you might have to also upload the updated expected .seq files with new raster times.

mcencini:
The naming looks fine to me as well @schuenke.
I also agree with removing the subsection, and with using `np.deg2rad` instead of manual `* np.pi / 180.0`
Perhaps we could enforce floating point values for hardcoded params (`rf_spoiling_inc = 117.0` etc).
Regarding vendor compatibility, perfect, let's defer it to a separate issue - perhaps we could consider simply exposing main system params in `main()` signature (raster times, max_grad, max_slew) as we are editing it in this current PR anyway, but leaving this to another PR sounds good too.

mcencini:
Ok, this seems a significant refactor effort. I would limit this PR to basic renaming of the variables (not of the events) and open a series of sub-issues to refactor individual sequences in separate PRs.

- **EPI**: they seem not to use ramp-sampling, maybe we could align with the upstream MATLAB [version](https://github.com/pulseq/pulseq/blob/master/matlab/demoSeq/writeEpiRS.m) during refactoring (can be a separate function).
- **TSE/HASTE**: they seem largerly overlapping, maybe better to double check if they are effectively the same sequence and remove one of the two (the other can be left as an alias for the surviving sequence). We should refactor to make use as much as possible of relevant toolbox functions (`pypulseq.calc_duration()`, `pypulseq.align()`, `pypulseq.split_gradient()`, `pypulseq.split_gradient_at(`)).
- Similar, `write_mprage` and `write_3Dt1_mprage` - if they are not equivalent (e.g., naive vs optimized) we should better document it.

schuenke:
> Ok, this seems a significant refactor effort. I would limit this PR to basic renaming of the variables (not of the events) and open a series of sub-issues to refactor individual sequences in separate PRs.
> 
>     * **EPI**: they seem not to use ramp-sampling, maybe we could align with the upstream MATLAB [version](https://github.com/pulseq/pulseq/blob/master/matlab/demoSeq/writeEpiRS.m) during refactoring (can be a separate function).
> 
>     * **TSE/HASTE**: they seem largerly overlapping, maybe better to double check if they are effectively the same sequence and remove one of the two (the other can be left as an alias for the surviving sequence). We should refactor to make use as much as possible of relevant toolbox functions (`pypulseq.calc_duration()`, `pypulseq.align()`, `pypulseq.split_gradient()`, `pypulseq.split_gradient_at(`)).
> 
>     * Similar, `write_mprage` and `write_3Dt1_mprage` - if they are not equivalent (e.g., naive vs optimized) we should better document it.

Sounds good to me. 

Do we still want to do the obvious renaming (e.g. `t_ex` -> `rf_ex_duration`) in this PR or do we only care about the function arguments here?

mcencini (commented):
I only left a small comment about common variable names but it otherwise look like a very good start to me, it should be ok for this PR. I also agree it can stay in `examples/scripts` subfolder for the time being and we'll move it to `doc` (Guide for Developers section?) when we refactor it later.

mcencini (approved):
Maybe I misunderstand @mzaiss , but it looks like we already accept tuples with length = 3 `(fov_x, fov_y, fov_z)`. Thank you very much @schuenke, LGTM!

mcencini:
> > Maybe I misunderstand @mzaiss , but it looks like we already accept tuples with length = 3 `(fov_x, fov_y, fov_z)`. Thank you very much @schuenke, LGTM!
> 
> Yeah, for the 3D sequences I changed the code to allow either a single float value (for isotropic FoV), or a 3-tuple with (fov_x, fov_y, fov_z). For (multi-) 2D sequences, I sticked with a 2-tuple (fov_x, fov_y) and additional `n_slices` and `slice_thickness` parameters, because I didn't want to change the sequence logic in this PR.
> 
> I agree that we could also allow 3-tuple for 2D sequences in the future, however, this will raise additional questions... e.g. how to behave if a 3-tuple `fov`, `n_slices` and `slice_thickness` is given, but `n_slices * slice_thickness != fov_z`. Would we introduce gaps if `fov_z > n_slices * slice_thickness`? What if `fov_z < n_slices * slice_thickness`?

Yeah, slice gaps are the reason why I think for 2D it is better to keep 2-tuple fov plus n_slices and thickness (+ spacing / gaps). I'd say for the moment this is a good start.

PR #367, opened 2026-02-17
Add PyPulseq tutorial(s)

schuenke:
In #366 , @mcencini came up with the idea to add different tutorials to the PyPulseq GitHub repo to provide a better starting point for new (Py)Pulseq users compared to the existing example sequences.

I picked up this idea and drafted the first notebook about the basics of PyPulseq of such a tutorial series. 

My idea would be to use this PR to discuss the different aspects we want to cover in this tutorials, how we want to structure it etc.

Furthermore, I already added `nbstripout` to our pre-commit config to ensure that all notebooks are pushed without outputs and clean execution counts etc. 

---
The current idea (taking feedback from @mcencini into account) would be to have the following structure:

- [x] Introduction / Basics
   - [x] (Py)Pulseq Intro
   - [x] Installation
   - [x] How to Import 
   - [x] `Opts` and `Sequence objects`
   - [x] Creating basic events (RF. ADC. delay)
   - [x] Building a sequence (`add_block()`)
   - [x] Checking, Plotting, Exporting
- [x] RF pulses
   - [x] shapes (sinc, gauss, bwt, apodization, ...)
   - [ ] make_arbitrary_rf() example
   - [x] slice-selection
   - [x] flip angle, phase offset, freq offset
   - [x] use cases: excitation, inversion, preparation, ...
- [ ] Gradients
   - [ ] grad channels
   - [ ] key concepts (rise/flat/fall time, slew rate...)
   - [ ] trapezoid gradients
   - [ ] extended trapz
   - [ ] arbitrary
   - [ ] applications (encoding, rephasing, spoiler)
- [ ] ADC and k-space readout
   - [ ] relationships (FoV, delta_k, matrix size...)
   - [ ] readout grad + adc alignment
   - [ ] k-space trajectory calculation
- [ ] Building a simple spoiled 2D gre sequence with fixed TE/TR
- [ ] Extend GRE sequence to allow for variable TE/TR
   - [ ] introduce more advanced timing calculations 
- [ ] Time-optimized GRE sequence
   - [ ] use extended trapz
- [ ] Extend GRE sequence to multi 2D/3D
- [ ] Add labels to the GRE sequence
- [ ] Make TE/TR soft delays
- [ ] Convert cartesian GRE to radial sequence
   - [ ] introduce rotate functions 
- [ ] Convert caetesian GRE to EPI
   - [ ] gradient surgery (split_gradient(), split_gradient_at(), add_gradients() and align())
   - [ ] (essentially, following [this](https://github.com/pulseq/tutorials/tree/main/11_from_GRE_to_EPI)).
- [ ] bSSFP
   - [ ] control flags (e.g. `once`) (see [here](https://github.com/pulseq/pulseq/blob/master/matlab/demoSeq/writeTrufi.m)). 
- [ ] ...

Looking forward to hear you feedback!


schuenke:
> Sure, that works! So perhaps for the dedicated raster time notebook we can follow the [shapes_and_times](https://github.com/pulseq/pulseq/blob/master/doc/pulseq_shapes_and_times.pdf) appendix to the official Pulseq specification, and that could be a good spot to discuss how to make sure the same sequence could be faithfully played on scanners from different vendors, as raster times are the main difference in this regard.
> 
> A few topics that I think would deserve some tutorials (I would say at least after labels):
> 
>     * from naive GRE (spoiler in a independent block) to optimized (readout+spoil in a single block): it represent the simplest applicationof extended trapezoid.
> 
>     * from GRE to EPI: in this notebook we could start from GRE, extend to multi-echo GRE and add blips to achieve EPI (essentially, following [this](https://github.com/pulseq/tutorials/tree/main/11_from_GRE_to_EPI)). This represent a good introduction to gradient surgery, including `split_gradient()`, `split_gradient_at()`, `add_gradients()` and `align()`. As an alternative, a deep dive into `FSE` could also work, but maybe EPI is more indicated.
> 
>     * bSSFP: this could be a good practical application of control flags, namely `once` (see [here](https://github.com/pulseq/pulseq/blob/master/matlab/demoSeq/writeTrufi.m)).
> 
>     * cartesian GRE to radial: here we can show the rotation extension (as soon as `v1.5.1+` is ready)

Thanks for the suggestions. I added an updated outline in my first post.

mcencini:
The RF example is also very nice - should we add a small subsection to explain `make_arbitrary_rf()`?

mcencini:
I would say either we keep it very simple, e.g., we implement a block or sinc pulse as an arbitrary rf pulse, or we may adapt (and reference) an existing tutorial like e.g., [sigpy.rf](https://github.com/jonbmartin/sigpyrf-demos-ISMRM-2021/blob/main/Ex2-Community-Tool-Design.ipynb) (starting from step 7). I would maybe opt for option 1, as the RF tutorial is intended to be very basic.

PR #370, opened 2026-02-21
Test sequence runnable script

fsantini:
This PR adds a runnable script "ppseqtest" that can be used to plot an existing .seq file. It is useful to visualize a third-party file, or generated with another framework.

A section "command-line utilities" was added to the README

Minor change:
- Links in the README were fixed and reorganized.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,9 @@
 3. [⚡ Lightning-start - PyPulseq in your browser!][section-lightning-start]
 4. [🏃‍♂ Quickstart - example scripts][section-quickstart-examples]
 5. [🤿 Deep dive - custom pulse sequences][section-deep-dive]
-6. [👥 Contributing and Community guidelines][section-contributing]
-7. [📖 References][section-references]
+6. [🖵 Command-line utilities][section-cli]
+7. [👥 Contributing and Community guidelines][section-contributing]
+8. [📖 References][section-references]
 
 ---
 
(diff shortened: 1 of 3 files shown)


mzaiss:
This is a helpful thing, we use similar scripts.

We realized that this is actually a "pulseq interpreter"

as it mimicks the interpreter on a real scanner.
 just an idea to maybe actually call such a script like that.

you could also make it similar to the new example sequences, regarding input, output (seq object) and a check for fov definition, which would turn this into a test very colse to a  scanner interpreter.

returning seq also allows then to run similar functions as for other  write_seq files afterwards.

like seq.plot or seq.calulate_kspace

fsantini:
Hi Moritz, thanks for the comment. I'm not sure I understand it though, as the script does very little except loading an existing sequence file, so what's the use of returning a seq object? It would be basically equivalent to do read().

schuenke:
I don't really get why I would need this to be honest.

I would still need to install Python and PyPulseq and the only advantage would be to be able to run a plot/check without starting my IDE and manually call seq.read() and seq.plot(), right?

However, I am not able to specify my system limits which makes timing checks and the test report a bit meaningless, no?

Maybe you can explain again when/where/why you would use this and maybe @mzaiss can also quickly explain why this should be seen as kind of a "virtual interpreter" !?

fsantini:
Hi Patrick, it is useful if you just want to use pypulseq as a utlity to look at existing seq files. In my specific case, I would generate seq files with nimpulseq, which doesn't have plotting capabilities, so I would then use this utility to look at the result. However, I understand that this might not be the majority use case for pypulseq users, so I'm happy to keep this (and other future command-line utilities) in a separate package that simply depends on pypulseq.

mzaiss:
for us its nice to have something generic that returns seq objects.

then i can use e.g. mrseq seqs  and files 
analogously.

in that sense mrseq is a built-in sequence, and such a function would be a pulseq interpreter for files.

but this does not need kspqce and other plotting.
and it must return seq objects

fsantini:
> for us its nice to have something generic that returns seq objects.
> but this does not need kspqce and other plotting. and it must return seq objects

Isn't this what Sequence.read() does?

mzaiss:
yes, but you need to define a seq first. so its a nice mini helper function, that coukd be extended  to limit tests that also run on a real scanner. that's what i thought you had in mind. but i take all my comments from this thread now back as i think im confusing everybody. please decide not based on my comments.

PR #375, opened 2026-03-01
Fix RF dead-time handling for slice-selective RF pulses (sinc/gauss)

dabosch:
Fixes a bug where RF delay is increased even when ramp-up time of the slice selection gradient gives sufficient delay.

Intended behavior: increase RF delay iff RF pulse delay is smaller than dead time.
Old actual behavior: modify RF delay if RF pulse or slice selection gradient has delay smaller than dead time.
In other words, the ramp-up time of the slice selection was not taken into account when checking if RF delay is sufficient.

Fixed by moving dead time check to a later position in the code, after slice selection gradient is generated and RF pulse is positioned correctly. If required, gz delay is adjusted as well.


schuenke (approved):
The old implementation actually did **NOT** lead to wrong delays, but the warning was wrong or at least confusing, because it could have happened that the rf.delay was further adjusted after the warning stated it was increase to the RF dead time.

If `return_gz=True`, we were/are anyway adjusting the delays silently, so that I decided to emit a warning only when `return_gz=False`, because there users should not expect any delay adjustments. 

Additionally, I applied the same changes to `make_gauss_pulse` and added/modified the tests for all make RF functions.

PR #378, opened 2026-03-09
Add rotate3D implementation

Moji3h:


mcencini (changes requested):
Thanks @Moji3h for this PR! I left some comments about styling and a suggestion to avoid a custom quat_to_matrix function.

In addition, I think we should add a test suite for this.

Moji3h:
Thanks for the comments. I updated it accordingly and resolved the issues.

src/pypulseq/rotate3D.py
```diff
@@ -0,0 +1,254 @@
...
+    rotation: Union[Any, None] = None,
+    rotation_matrix: Union[np.ndarray, None] = None,
+    system: Union[Opts, None] = None,
+) -> List[SimpleNamespace]:
+    """
+    Rotate gradient events by a 3*3 rotation matrix or quaternion.
+
+    Non-gradient events (e.g. RF, ADC, delay) are passed through unchanged.
+
+    Parameters
```
mcencini, line 129: There are missing dashes for correct rendering:

```
Parameters
------------
```

same for the other sections of the docstring

src/pypulseq/rotate3D.py
```diff
@@ -0,0 +1,254 @@
...
+            scaled = scale_grad(grad=events_to_rotate[ax_in], scale=w)
+            scaled.channel = axes[j]
+
+            if grad_out_curr is None:
+                grad_out_curr = scaled
+            else:
+                # grad_out_curr = align_gradient_to_raster(grad_out_curr, system.grad_raster_time)
+                # scaled = align_gradient_to_raster(scaled, system.grad_raster_time)
+                grad_out_curr = add_gradients([grad_out_curr, scaled], system=system)
+                grad_out_curr = align_gradient_to_raster(grad_out_curr, system.grad_raster_time)
```
mcencini, line 249: @Moji3h, do you have any comment regarding @fzimmermann89 concern?

mcencini:
@Moji3h thank you for the review! I still see some issues with the docstring, as well as the remaining issue of post rotation gradient alignment.

In addition, I would strongly advise to add some unit tests, it is important not to decrease the coverage. Thanks again for your support!

src/pypulseq/rotate3D.py
Moji3h: I had some errors related to this line in the past, but I couldn’t find any examples showing the error.

PR #380, opened 2026-03-18
Add rf_plot modes (auto/abs/real/imag) to sequence plotting

JosephGWoods:
Users can now choose rf_plot = 'auto', 'abs', 'real', 'imag' (default = 'auto' preserves existing behaviour).

If seq.plot(rf_plot = 'auto'), the RF magnitude is plotted in the RF magnitude subplot, unless RF pulses are purely real or imaginary, in which case the real or imaginary part is plotted and the phase waveform is adjusted accordingly.
If seq.plot(rf_plot = 'abs'), the RF magnitude is always plotted.
If seq.plot(rf_plot = 'real') or seq.plot(rf_plot = 'imag'), the real or imaginary parts are plotted.

Additional tests have been added to tests/test_sequence_plot.py to test this new functionality.

### DIFF
--- a/src/pypulseq/Sequence/sequence.py
+++ b/src/pypulseq/Sequence/sequence.py
@@ -2,7 +2,7 @@ import math
 from collections import OrderedDict
 from copy import deepcopy
 from types import SimpleNamespace
-from typing import Any, List, Tuple, Union, Literal
+from typing import Any, List, Literal, Tuple, Union
 from warnings import warn
 
 try:
--- a/src/pypulseq/Sequence/sequence.py
+++ b/src/pypulseq/Sequence/sequence.py
@@ -1057,6 +1057,7 @@ class Sequence:
         overlay: SeqPlot = None,
         stacked: bool = False,
         show_guides: bool = False,
+        rf_plot: str = 'auto',
     ) -> SeqPlot:
         """
         Plot `Sequence`.
@@ -1090,6 +1091,11 @@ class Sequence:
             If False, use separate figures for RF/ADC and gradients.
         show_guides : bool, default=False
             If True, enable dynamic vertical hairline guides that follow the cursor. Requires `mplcursors`.
+        rf_plot : {'auto', 'abs', 'real', 'imag'}, default='auto'
+            Determines how to plot RF waveforms in the RF magnitude plot.
+            If 'auto', plots magnitude for all RF events except those that are purely real or imaginary, which are plotted as such.
+            If 'abs', plots magnitude for all RF events.
+            If 'real' or 'imag', plots the respective component for all RF events.
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


src/pypulseq/Sequence/sequence.py
```diff
@@ -1057,6 +1057,7 @@ def plot(
         overlay: SeqPlot = None,
         stacked: bool = False,
         show_guides: bool = False,
+        rf_plot: str = 'auto',
```
schuenke: Could improve typing using 
`rf_plot: Literal["auto", "abs", "real", "imag"] = "auto"`

tests/test_sequence_plot.py
```diff
@@ -72,3 +72,62 @@ def test_plot_stacked_behavior():
...
+        seq.plot(plot_now=False, rf_plot='invalid')
+        raise AssertionError('Should have raised ValueError for invalid rf_plot mode')
+    except ValueError as e:
+        assert 'Unsupported RF plot type' in str(e)
+
+    plt.close('all')
+
+
+def test_plot_rf_plot_auto_default():
+    """Test that rf_plot='auto' is the default and produces same result as no argument."""
```
schuenke, line 105: This only checks that both calls return a figure, but not that they yield the same result. I am fine with either adjusting the test to really compare the output, but also with just adjusting the docstring.

src/pypulseq/utils/seq_plot.py
schuenke: We should probably adjust the label to cover `RF mag` , `RF real`, `RF imag` and something like `RF (auto: abs/real/imag)`. For auto, it's a bit tricky because we can potentially have a real only and an imag only RF in the same plot. We could ofc track what RF types are included in the plot, but labeling it as just `RF (auto: abs/real/imag)` would be fine for me.

schuenke (changes requested):

src/pypulseq/Sequence/sequence.py
JosephGWoods: Good idea. Done: 68cda874e90c6e74a95c778488a9e679e7be96c5

src/pypulseq/utils/seq_plot.py
JosephGWoods: I have updated the ylabel now to cover the different cases and used your suggestion for auto.

tests/test_sequence_plot.py
JosephGWoods: Yes, good spot. I have updated the unit test to check that the line data is identical for both figures. See c9dc0ac64a1b84840be295d6ae179117879d5759

schuenke (approved):
Thanks for tackling my review comments. 

I did some minor style adjustments and corrections in [cad2f37](https://github.com/imr-framework/pypulseq/pull/380/commits/cad2f37329a4339469d439ff5b3bbaf4959a22bd) so IMO it looks good now.

I will do the merge soon.

PR #385, opened 2026-04-27
Update safety PNS code and add chronaxie/rheobase Python port

LeoMcBills:
- Updates `safe_pns.py` with the latest safety branch changes.
- Adds `src/pypulseq/safety/pns/chronaxie_rheobase.py`.
- Keeps PNS safety tooling in Python and ready for integration/testing in the main code path.


schuenke:
Wouldn't it make more sense to keep this as a separate package, which we add as an additional dependency so that we can call the functions in `Sequence.calc_pns()` etc

PR #388, opened 2026-05-15
Allow None inputs in calc_duration function

schuenke:
This PR adds handling of `None` to the `calc_duration` function.

This allows code blocks like

```Python
# Passing an optional event (e.g. a gradient that may or may not exist):

optional_event = optional_object if use_optional_event else None
block_duration = pp.calc_duration(any_existing_event, optional_event)
```

If `None` is the only input argument, a `ValueError` will be raised. 

I also improved and updated the documentation and added additional tests.


src/pypulseq/calc_duration.py
```diff
@@ -3,31 +3,45 @@
 from pypulseq.block_to_events import block_to_events
 
 
-def calc_duration(*args: SimpleNamespace) -> float:
+def calc_duration(*args: SimpleNamespace | dict | float | int | None) -> float:
```
mcencini: I absolutely prefer the above type hints styling, but I am not sure whether is it better to keep the current pypulseq Union[...] type hints style and defer the refactor of type hints styling to a separate dedicated PR. What do you think?

src/pypulseq/calc_duration.py
```diff
@@ -3,31 +3,45 @@
...
+    durations (i.e. the duration of a block comprised of those events).
+
+    `None` inputs are ignored, which allows optional events to be passed without an explicit
+    conditional. If `None` is the only provided input, a `ValueError` is raised.
 
     Parameters
     ----------
-    args : SimpleNamespace
-        Block or events.
+    args : SimpleNamespace | dict | float | int | None
```
mcencini: Same consideration as above

mcencini (commented):
Apologies for the delay @schuenke, I was traveling and just got back. I have just a couple of notes. RE None handling: I think it makes sense to return duration 0.0 for None events.

src/pypulseq/calc_duration.py
schuenke: good catch. Adjusted in [2c4853e](https://github.com/imr-framework/pypulseq/pull/388/commits/2c4853ed853f5acbb3c17b469342ef6b36e79a8b)

mcencini (approved):
Looks good!

PR #389, opened 2026-05-15
Allow None inputs in Sequence.add_block method

schuenke:
Similar to #388 , this PR allows to have `None` inputs in the `Sequence.add_block()` method. 

For the example from #388 :

```Python
# Passing an optional event (e.g. a gradient that may or may not exist):

optional_event = optional_object if use_optional_event else None
block_duration = pp.calc_duration(any_existing_event, optional_event)
```

this PR will allow to use a simple code snippet like:

```Python
seq.add_block(any_existing_event, optional_event)
```

instead of

```Python
if optional_event is not None:
    seq.add_block(any_existing_event, optional_event)
else:
    seq.add_block(any_existing_event)
```


src/pypulseq/Sequence/sequence.py
```diff
@@ -167,7 +167,7 @@ def adc_times(self, time_range: Union[List[float], None] = None) -> Tuple[np.nda
 
         return t_adc, fp_adc
 
-    def add_block(self, *args: SimpleNamespace) -> None:
+    def add_block(self, *args: SimpleNamespace | None) -> None:
```
mcencini: Same as #388: should we keep `Union[SimpleNamespace, None]` for now and defer styling refactor to a separate PR?

schuenke: changed it to the typical PyPulseq style. Again, good catch 👍

mcencini (approved):

PR #390, opened 2026-05-15
Fix publish-to-pypi config / workflow

schuenke:
The 2nd nested tags mapping was probably unintentional and leads to failing workflows... 
I think I introduced it myself in #355


mcencini (approved):
lgtm!

PR #392, opened 2026-05-22
Bug fix related to phase_offset and phase_ppm units

JosephGWoods:
1) In seq_plot.py, added the missing seq.system.gamma from the calculation of full_freq_offset and full_phase_offset from freq_ppm and phase_ppm, respectively.
2) Corrected the issue of doc-strings incorrectly reporting the phase_offset of RF and adc objects being in Hz. The correct units are rad.
3) Updated other missing unit information in the doc-strings.


mcencini (approved):
lgtm! and many thanks for updating the phyisical units in docstring, I think it is very helpful for the users

PR #396, opened 2026-05-27
Update Python versions

### LINKED ISSUE #393: Inconsistent Python Version
**Describe the bug**
The supported Python version range is currently inconsistent across the project.
- `README.md` references to Python `3.7--3.13`
- `pyproject.toml` references to Python `>=3.8`
- `.github/workflows/pytest.yml` tests references to Python `3.8` through `3.12`

**To Reproduce**
Refer screenshots

**Expected behavior**
This makes it unclear which Python versions are actually supported. 
Users may assume Python 3.7 or 3.13 are supported based on the README.md , while workflows and pyproject.toml do not fully confirm that. Clear range for Python version should be maintained across all the references.

**Screenshots**

Image

Image

Image

**Desktop (please complete the following information):**
Not Applicable

**Additional context**
Recommended Fix
- Align Python version references
- If Python 3.13 support is intended, add it to pyproject.toml
- If Python 3.7 is no longer supported, remove it from the README.md.
- Consider raising the minimum to `>=3.9` or `>=3.10` if legacy support is not essential.

schuenke:
Fixes #393 

- use Python 3.14 instead of 3.12 as default version for tests and test reports
- reduce test matrix to current minium (3.8) and maximum (3.14) Python version
- update python badge in readme

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.7", "3.12", "3.13", "3.14"]
         include-sigpy: [true]
 
     steps:
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.12", "3.13", "3.14"]
+        python-version: ["3.8", "3.14"]
         include-sigpy: [true]
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/pytest.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.14"]
```
conveyanand, line 20: The `python-version` defines `Python 3.8 and 3.14` jobs, but the workflow did not use version for python setup.
The current implementation will run both jobs using the default Python from `ubuntu-latest`.
For example, if the current `ubuntu-latest` image version is 24.04, the default python version is Python 3.12.

schuenke:
I would like to (try to) release a new version (1.5.0.post2) after we moved the repo to the pulseq organisation and merge this before the release. 

IMO, testing only the minimum and (current) maximum Python version is fine, but we could also switch to minium + 3 latest if wanted. Let me know what you think @mcencini @sairamgeethanath 

@sairamgeethanath Just a quick reminder to check the PyPi owner rights

mcencini (approved):
I think testing min and max is perfectly fine.

PR #402, opened 2026-06-10
Fix scanner signature error and add binary signature verification

albertx-wang:
* `read_binary.py` and `write_binary.py`:
  Added binary `.bseq` read/write support with MD5 signature serialization and deserialization.

* `read_seq.py` and `write_seq.py`:
  Updated text `.seq` signature handling. `write_seq.py` now computes the MD5 hash from raw file bytes before appending `[SIGNATURE]`, and `read_seq.py` restores the signature metadata consistently when reading signed text files.

* `verify_file_signature.py`:
  Added a helper that recomputes the MD5 hash over the signed portion of text or binary Pulseq files and compares it with the stored signature, matching the MD5 verification behavior of MATLAB `verifyFileSignature.m`.

* `sequence.py`:
  Exposed `read_binary()` and `write_binary()` on `Sequence`, and added shared binary section codes used by binary IO and signature verification.

* `test_binary_signature.py`:
  Added binary signature roundtrip coverage for `write_binary()`, `read_binary()`, and `verify_file_signature()`, matching the binary signature roundtrip covered by MATLAB `testBinarySignature.m`.

### DIFF
--- a/src/pypulseq/Sequence/read_binary.py
+++ b/src/pypulseq/Sequence/read_binary.py
@@ -283,4 +283,3 @@ def _read_soft_delays(fid, seq):
             seq.soft_delay_hints2.append(hint)
         seq.soft_delay_hints[hint] = int(num)
         seq.soft_delay_library.insert(event_id, np.array([num, offset, factor, hint_id], dtype=float))
-
--- a/src/pypulseq/Sequence/read_binary.py
+++ b/src/pypulseq/Sequence/read_binary.py
@@ -1,0 +1,286 @@
+from __future__ import annotations
+
+from collections import OrderedDict
+
+import numpy as np
+
+from pypulseq.event_lib import EventLibrary
+
+
+def read_binary(self, filename: str) -> None:
+    codes = self.get_binary_codes()
+    with open(filename, 'rb') as fid:
+        magic_num = _read_scalar(fid, np.int64)
+        if magic_num != codes['fileHeader']:
+            raise ValueError('Not a Pulseq binary file')
+        version_major = _read_scalar(fid, np.int64)
+        version_minor = _read_scalar(fid, np.int64)
+        version_revision = _read_scalar(fid, np.int64)
+        if version_major != int(self.version_major):
+            raise ValueError(f'Unsupported version_major {version_major}')
+        if version_minor != int(self.version_minor):
+            raise ValueError(f'Unsupported version_minor {version_minor}')
+        if version_revision != int(self.version_revision):
+            raise ValueError(f'Unsupported version_revision {version_revision}')
+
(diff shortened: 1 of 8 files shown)


src/pypulseq/Sequence/read_binary.py
```diff
@@ -0,0 +1,285 @@
...
+def _read_soft_delays(fid, seq):
+    num_events = int(_read_scalar(fid, np.int64))
+    for _ in range(num_events):
+        event_id = int(_read_scalar(fid, np.int32))
+        num = float(_read_scalar(fid, np.int32))
+        offset = float(_read_scalar(fid, np.int64)) * 1e-12
+        factor = float(_read_scalar(fid, np.float64))
+        hint_len = int(_read_scalar(fid, np.int32))
+        hint = fid.read(hint_len).decode('ascii')
+        if hint in seq.soft_delay_hint_ids:
```
btasdelen, line 278: Is there a reason hint ID logic here is different from the `read_seq()` and requires extra variables stored?

src/pypulseq/Sequence/read_seq.py
```diff
@@ -97,7 +97,7 @@ def read(self, path: str, detect_rf_use: Union[bool, None] = None, remove_duplic
                 self.signature_type = temp_sign_defs['Type']
             if 'Hash' in temp_sign_defs:
                 self.signature_value = temp_sign_defs['Hash']
-                self.signature_file = 'Text'
+                self.signature_file = 'text'
```
btasdelen, line 100: I don't know if any code out there uses this variable, but upstream Pulseq kept this as 'Text', so might be a better idea to keep it as 'Text', even though it is inconsistent capitalizaiton with 'bin'

src/pypulseq/Sequence/read_binary.py
albertx-wang: The difference was to align with MATLAB’s binary read/write behavior. To avoid adding complexity here, I’ll move this into a separate PR.

src/pypulseq/Sequence/read_seq.py
albertx-wang: Thanks for catching this, I will keep it as Text for consistency!

src/pypulseq/Sequence/read_binary.py
btasdelen: I see. It was a conscious decision on our part to slightly deviate from Matlab Pulseq's logic. You can see the relevant discussion here #264. I think it is best to mimic `read_seq()` function to avoid any behavioral differences.

albertx-wang: Got it!

PR #417, opened 2026-08-17
Improve gradient raster time accuracy for split_gradient_at function

ckolbPTB:
Some vendors use sub mu second gradient raster times. This leads to "not on raster" errors with the current code. This PR fixes this by increasing the accuracy of the rounding carried out in `split_gradient_at`


schuenke (approved):
Looks like a save fix that does not effect standard non-sub-µs raster times.

